PR #18, opened 2017-04-13
Update Connections API support

leonardt:
Main internal change to module definitions:
```diff
-    unordered_set<Connection> connections;
+    unordered_set<Connection*> connections;
```

This makes it easier for the API implementation (passing around pointers to connections rather than having to manage the connection objects themselves). I updated the typechecking and fileio code that uses connections to reflect this change (mainly `.` to `->`).

Does this make sense or does this violate a design choice?

### DIFF
--- a/TODO
+++ b/TODO
@@ -38,10 +38,6 @@ Known Bugs
   Person: Ross
 
 Lenny TODO
-  Add Following functions to C interface and Python interface
-    DONE ModuleDef.getConnections() //Get all the Wireable* Wireable* pairs 
-        TODO test (Blocked until Ross fleshes out problems with ModuleDef.connections)
-
   Maybe other python convenience functions for traversing the graph. We need to make sure that Caleb can use these functions.
 
   We need to figure out how to nicely pass errors to python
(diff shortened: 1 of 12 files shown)


rdaly525:
@leonardt  I think this changes some bit of functionality that I would like to have in the connnections. When I call wire, I do not want to save a connection if it already exists. This is why I had the guaranteed  ordering and placing in an unorderd_set rather than a vector. Your change does not preserve this because now the unordered_set is trying to do key hash comparisons based on an arbitrary pointer value. 

In terms of the C api, you could either
  create a new list of Connection*s in the coreir.cpp:BlahGetConnections(...)
  And just memory manage each new connection in Context,

 OR

  I think an easier solution, is since Connection is so simple, (just a c++ pair) explicitly define COREIRConnection to be a struct with
  COREIRWireable* first;
  COREIRWireable* second;

This would allow you to get rid of the accessor functions

leonardt:
@rdaly525 updated the changes to revert the implementation of `module_def.connections` and instead use a struct wrapper

```
typedef struct COREConnection  {
     COREWireable* first;
     COREWireable* second;
 } COREConnection;
```

PR #22, opened 2017-04-20
Revert "Split stdlib.hpp into .cpp/.hpp"

rdaly525:
Reverts rdaly525/coreir#20

I want to keep the libs separate from the src. Src should be for building the core CoreIR functionality, whereas libs should be for libraries implemented in coreIR. Can you just update the makefile in libs?


jeffsetter:
Please consider the newest pull request over reverting back.

PR #25, opened 2017-04-20
Order of argument passing is incorrect in ModuleDef::addInstance

THofstee:
Adding an instance of a Generator has genargs and config arguments swapped in the function call.


THofstee:
Related line:
`Instance* ModuleDef::addInstance(string instname,Generator* gen, Args genargs,Args config) {`

THofstee:
Line 170 of fileIO.cpp confirms this is the right order so I'm just going to merge this.
`mdef->addInstance(instname,gref,genargs,config);`

rdaly525:
Thank you for catching that, but in the future please do not merge your own pull requests to master. I will be creating a dev branch so that master will always work (even with other projects at stanford)

Issue #27, opened 2017-04-22
Generator Params vs. TypeGen Params


THofstee:
Currently, the Generator constructor requires that the Params provided to the Generator match the Params provided to the TypeGen. There are cases where you would like the Generator to take additional parameters over the TypeGen. For example, consider a line buffer.

The line buffer needs to know at least 5 things: the width and height of the taps you want, the width of the image, the height of the image, and the bits per pixel. On the other hand, the type of the line buffer is very easily represented, it is just a single pixel in, and width*height pixels out. The TypeGen does not require any knowledge about the width and height of the image.

Instead, I propose that the Parameters provided to the Generator be a superset of the Parameters provided to the TypeGen.

THofstee:
An argument against would be we would need to know the image width in order to know the delay of a module, but I'm going to disagree with that by saying the delay of a module should be a property of a module, not a property of its interface.

I'm open to hear what anyone else thinks.

rdaly525:
I think this is good and allows typegens to be able to be used by a larger set of generators.

PR #28, opened 2017-04-22
Require that generator params/args be a superset of typegen params

THofstee:
This is a fix for issue #27.

### DIFF
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -35,11 +35,11 @@ bool operator==(const Args& l, const Args& r) {
 }
 
 bool checkParams(Args args, Params params) {
-  if (args.size() != params.size()) return false;
+  if (args.size() < params.size()) return false;
   for (auto parammap : params) {
     auto arg = args.find(parammap.first);
     if (arg == args.end()) return false;
-    if (!arg->second->isKind(parammap.second) ) return false;
+    if (!arg->second->isKind(parammap.second)) return false;
   }
   return true;
 }
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -35,7 +35,7 @@ bool operator==(const Args& l, const Args& r) {
 }
 
 bool checkArgs(Args args, Params params) {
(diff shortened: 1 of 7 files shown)


THofstee:
Retargeted onto dev.

THofstee:
I think `checkArgs` might be a better name than `checkParams`, but I'm not sure if that's the best name for this.

THofstee:
Out of curiosity, what would be the reasoning for having another function test whether or not they are exactly equal?

rdaly525:
I do the matching check on the Args being passed into the generator. Is there a good situation where you would want to pass in more args to a generator instance than the generator needed? Perhaps?
I do totally agree that the typegen should be a subset of the generator args.

THofstee:
I'm not sure. Previously in the other branch I was working in I used it out of convenience because I didn't need a type to pass into runGenerator, but I would get a segfault if I passed in a nullptr so I just grabbed the TypeGen for the Generator and ran it with the arguments for the generator.

I think it's probably better to only have the equality check instead of subset when you provide arguments to the parameters. I can't think of any programming languages that allow you to pass in more arguments to a function than the function has parameters, and I can imagine a lot of nasty bugs that might result from this.

I'll need to double check to see if the nullptr passed to runGenerator still causes a segfault in the dev branch. I'll clean this up a bit tomorrow.

rdaly525:
Just as a warning, I did clean up the infrastructure for running the typegen and generator functions a bit. I made them into an actual method of the class rather than a random function (although I did keep that ability to pass in a random function). If the API is not clear, definitely let me know and add it to Issue #33.  Also if you are getting segfaults, then definitely report an issue. Ideally you should not be able to segfault.

rdaly525:
Thanks again for looking into this stuff! I know it is annoying that things keeps changing, but that is the definition of Agile!

THofstee:
Seeing as how there is only one function now to check that args exactly match params, I've just left it renamed `checkArgs` similar to other compilers.

PR #29, opened 2017-04-24
Update shared library path for Python bindings

leonardt:
It looks like libreorg is not buildling on the CGRAFlow branch, so I will defer the integration test of this branch to the eventual libreorg pull request.

### DIFF
--- a/bindings/python/coreir/__init__.py
+++ b/bindings/python/coreir/__init__.py
@@ -15,7 +15,7 @@ def load_shared_lib():
 
     dir_path = os.path.dirname(os.path.realpath(__file__))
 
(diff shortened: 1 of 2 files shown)


leonardt:
Build now passes on linux with a reordering of the linker flags based on the documentation here:
https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html

Specifically,
> It makes a difference where in the command you write this option; the linker searches and processes libraries and object files in the order they are specified. Thus, ‘foo.o -lz bar.o’ searches library ‘z’ after file foo.o but before bar.o. If bar.o refers to functions in ‘z’, those functions may not be loaded.

Thus, we link with `-lcoreir` after the object files

Issue #31, opened 2017-04-25
Update Generators and TypeGens to use a method called 'run' instead of a function pointer


rdaly525:
Currently I have an ugly function pointer attached to the generators and Typegens. This should definitely just be a normal method.

One way to maintain usage of passing in a function pointer is to allow these classes to be inherited from the base class like:

class PassFunPtr : Typegen {
  FunPtrType f;
  PassFunPtr(FunPtrType f) : f(f) {}
  Type* run(...) {
     f(...)
  }
}

rdaly525:
This is fixed in Genupdate

PR #36, opened 2017-04-27
Genupdate

rdaly525:
I did an overhaul of the way to run typegen functions and generators in order to make it more extensible.

Typegens now have a getType(Args) method which will run a virtual function called computeType(). This function can be overloaded from an inherited class (from TypeGen). This is a little nicer because now you can create TypeGens without having to pass in an ugly function pointer. I still allow you to pass in a function pointer though for the constructor.

This is also what I did for Generators. Generators now have a pointer to a GeneratorDef class which has a single method which will run the generator for you. In the same way, you can create generators by inheriting from the class and overloading the virtual function

Other minor fixes:
TypeGens are passed via pointers instead of values
addDef ->setDef (naming consistency)
Added actual Test passing checking for the unit tests


rdaly525:
This should close #31

PR #48, opened 2017-05-02
Added simple array creation sugar

rdaly525:
Added simple api on types.

Type* t = c->BitIn();
Type* arr = t->Arr(width)->Arr(3)->Arr(3)


rdaly525:
@jeffsetter 
(Although the change is in dev, not master)

PR #54, opened 2017-05-04
API convenience added

rdaly525:
changed WirePath to SelectPath for better naming. 
changed ancestors to selectpath for better naming. 
changed SelectPath to just be a vector of strings instead of the weird pair thing. 
Add the ability to specify wiring and selects by a '.' deleminated naming convention (More compact to write!)
Updated the connection unit test to use this notation.

Also added an assertion on sel to prevent segfaults for bad select

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,12 @@ all:
 test:
 	make -C tests
 	cd tests; ./run
+	
+pytest:
+	pip install -e bindings/python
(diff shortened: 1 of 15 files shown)


rdaly525:
@leonardt Sorry, I want to use this code from dev. Feel free to still review if you want!

PR #67, opened 2017-05-13
Inline

rdaly525:
Changed API:
ModuleDef.wire -> ModuleDef.connect

Added to API:
ModuleDef.disconnect(...)
ModuleDef.removeInstance(...)
Instance.inline()

Added flatten() to passes

Updated addN test to use flatten/inline
updated connect test to use disconnect and removeInstance


rdaly525:
Should fix #66

PR #72, opened 2017-05-17
Dview

rdaly525:
Adding in the API for viewing/traversing modules as directed graphs.


rdaly525:
Thanks Lenny for all the work here!
@cdonovick, you should now be able to use the Directed Graph view to traverse modules easier.

PR #81, opened 2017-05-27
Linebuffer Generator and CoreIR graphviz viewer

jeffsetter:
Change1: Linebuffer is added to cgralib, so it can be instantiated in other designs

Change2: Added additional functionality to saveModule(). It now checks the filename extension where
    .json creates the json file as before
    .txt creates a dot file that can be viewed as a png with  "dot -Tpng [filename.txt] > graph.png"


rdaly525:
Just realized you cannot really mix undirected and directed. I think you should go with using the directed Module which will break down the entire module into directed components.

Issue #50, opened 2017-05-03
Create an Instance Pass framework


rdaly525:
This should be a framework for iterating over all instances in the design and performing some set of actions on the instance.

An initial step would basically implement 1 to 1 replacement. 
A second step would be able to swap out multiple things.

rdaly525:
The way that LLVM does this is they have all the 'Uses' of a function within a LLVM::Module, and you have a command that just says replace all instances with this other instance.

I could do the same thing for CoreIR where I basically keep pointers to all instances (within a namespace) which use a specific Module. Then add a function from module that says replace all instnaces of this module with other module.

rdaly525:
Should have a function like:

Instantiable* replaceAllUsesWith(Instantiable*)

rdaly525:
This was already put in

PR #84, opened 2017-05-31
Instpass

rdaly525:
High level overview is that I created a match and replace function for use in the mapper (and others)

This of course required a bunch of other changes in the design. I do not think there should be any API changes, but a couple of API additions including:

matchAndReplace(...)
addPassthrough(...)
Module.copy()
ModuleDef.copy()
... (probably more)

Most of the changes should be laid out in the commit messages. Note, I was stupid for the commit called "message". That commit description is actually in the last commit.


src/lib/passes/flatten.cpp
```diff
@@ -12,10 +12,11 @@ void flatten(Context* c, Module* m, bool* err) {
   bool inlined = true;
   while (inlined) {
     inlined = false;
+    //TODO Not sure why this is not segfaulting
```
rdaly525: At least I left a TODO! haha

rdaly525:
Should fix #50

Issue #79, opened 2017-05-25
Change all the get functions which return strings to return a reference to the string


rdaly525:

rdaly525:
Fixed from pull request #85

THofstee:
Just ran into a bug with `COREDirectedConnectionGetSrc/Snk`, all the locations where there is a `newConstStringArray` points into a `std::vector<std::string>` from the C++ API, which is freed since nothing holds on to that vector. Either the char** needs to use malloc for each entry, or the vector needs to be added to a global list like some of the other functions.

rdaly525:
@THofstee, Thanks I will try to fix this

rdaly525:
@THofstee Can you confirm #92 fixes this issue?

THofstee:
seems like it is fixed.

PR #85, opened 2017-06-01
String -> const string &

rdaly525:
Updated (hopefully) all functions which return a string to return a const string reference instead


rdaly525:
This should fix issue #79

PR #111, opened 2017-07-05
[python-bindings] refactor interface - Instance.get_config_value(key) -> Instance.config[key].value

### LINKED ISSUE #110: [python-bindings] Fix interface consistency issues with `get_config_value`
Make the interface consistent with `instance.generator_args`.
* `instance.get_config_value(name) -> instance.config[name]`
* Return an `Arg` object rather than the raw value

leonardt:
Makes the interface more pythonic and consistent with `Instance.generator_args`.  Uses the same C-api logic so this is purely a cosmetic change in the Python bindings.

@cdonovick The interface changes from `add8_inst.get_config_value("init")` to `add8_inst.config["init"].value` (note that indexing `instance.config` returns an `Arg` object which is unpacked with the `.value`)

Closes #110

### DIFF
--- a/bindings/python/coreir/wireable.py
+++ b/bindings/python/coreir/wireable.py
@@ -67,7 +67,7 @@ class InstanceConfigLazyDict:
 
 class Instance(Wireable):
     def __init__(self, ptr, context):
-        super().__init__(ptr, context)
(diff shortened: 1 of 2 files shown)


cdonovick:
Looks good just let me know when its merged to master

PR #112, opened 2017-07-06
Fix bad assert

cdonovick:
`assert(False,"NYI!")` always evaluate as `True` because  `bool((False,"NYI!"))==True`.
The statement you meant was
`assert False,"NYI!"`
However, why `assert False` when you can `raise NotImplementedError`?


leonardt:
@rdaly525 should we merge this to dev first or just go straight to master? Purely a style change to I think it could be okay to go directly to master, but it's up to you if you want everything flow through dev first, I'm indifferent

PR #114, opened 2017-07-08
Added default args for both genargs and configargs

rdaly525:
Also did a minor rename of Genarg -> GenArg, Configarg -> ConfigArg
Also prettier printing for modules

### DIFF
--- a/src/fileIO.cpp
+++ b/src/fileIO.cpp
@@ -21,15 +21,6 @@ Params json2Params(json j);
 Module* getModSymbol(Context* c, string nsname, string iname);
 Generator* getGenSymbol(Context* c, string nsname, string iname);
 
-//Helper function for when args is a subset of params
-Params extractParams(Args args, Params params) {
-  Params ret;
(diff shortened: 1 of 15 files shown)


src/fileIO.cpp
```diff
@@ -523,6 +540,11 @@ json Module::toJson() {
 json Generator::toJson() {
   json j = Instantiable::toJson();
   j["genparams"] = Params2Json(genparams);
+    cout << "AM I HERE0??" << endl;
```
leonardt: Seems like a debugging message, can we silence these with a comment? (Or perhaps add a debug_print macro that's off by default)

src/fileIO.cpp
```diff
@@ -523,6 +540,11 @@ json Module::toJson() {
...
+  if (!defaultGenArgs.empty()) {
+    cout << "AM I HERE1??" << endl;
```
leonardt: Same as above

PR #121, opened 2017-07-22
Implement addInstance for Instantiable* arguments

leonardt:


rdaly525:
What is the reason for adding this function? Presumably if you are knowingly passing in genargs, you know it should be a generator and could pass in the instantiable casted to a generator. My rule that I have been trying to follow has been that if the function takes in genargs, you should pass in the generator instead of the instantiable. Also I want to make sure that genargs and configargs are never confused especially when one or both has a default value.

I should write a more explicit spec on this, but casting/type checking should be really simple with the functions isa<>, cast<> and dyn_cast<>

So to check an Instantiable is a generator you can simply just do isa<Generator>(instatiable_pointer).
To do a cast you know will pass, use cast<Generator>(instantiable_pointer) 
To do a cast which might return null, use dyn_cast<Generator>(instantiable_pointer)

Note that you do **not** have to do isa<Generator*>, these functions infer pointers (and const) automatically.

leonardt:
That makes sense, I think I can implement the functionality I wanted in a slightly different way that doesn't require this so I'll close it for now

leonardt:
I wanted to be able to do
```
Instantiable* instantiable = namespace->getInstantiable(name);
module_def->addInstance(instance_name, instantiable, ...);
```
But now I realized that if it's a generator I'll need to include some generator arguments so I'll need to dispatch on the instantiable type before calling addInstance and create the genargs if needed. At that point I'll already have the type and can call the right addInstance function, so the overloaded version shouldn't be needed

PR #134, opened 2017-08-02
Initial Firrtl backend

rdaly525:
Firrtl backend does not yet support clock types and registers.
Also some fixes with the way that generators are handled.
changed getModuleDef -> getContainer
Added api to add and delete "ports" (record fields) from modules using an instance pass. Need some good tests for this. This will be a nice way to change the names of port modules, add in clocks and enables, and debug signals.


src/ir/moduledef.cpp
```diff
@@ -223,6 +223,20 @@ void ModuleDef::connect(std::initializer_list<const char*> pA, std::initializer_
...
+}
+
+Connection ModuleDef::getConnection(Wireable* a, Wireable* b) {
+  auto sorted = std::minmax(a,b);
+  Connection con(sorted.first,sorted.second);
+  if (connections.count(con) > 0) {
+    return *connections.find(con);
+  }
+  return con;
+}
```
leonardt: If I understand this correctly, it will return a new connection if it's not present in the ModuleDef, but won't add it to the `connections` data structure. That seems confusing, is there a reason for that?

The behavior I would expect is some sort of error or null value if the connection isn't present, but perhaps this "default" connection behavior is useful somewhere else?

rdaly525: Yeah good catch. This should be an assertion.

rdaly525: Actually now that I think of it it should return the equivalent of a nullptr instead of an assertion. I think that is how I have been doing has/get functions.

PR #139, opened 2017-08-09
Initial implementation of FIRRTL primitive wrapper generator

leonardt:
* Treats all inputs/outputs as UInts. It uses the `as{UInt,SInt}` ops to "cast" them before/after the primitive.
* These don't take into account the widths (i.e. the output of an `add` op is a different width than the coreir `add`). We can either try to insert that logic into the wrappers, or we can emit "select" operators around primitive ops to make sure the output widths match coreir semantics.

### DIFF
--- a/tools/gen_firrtl_modules.py
+++ b/tools/gen_firrtl_modules.py
@@ -1,0 +1,92 @@
+import itertools
+from collections import namedtuple
+
+
+class Source:
+    def __init__(self):
+        self._source = ""
+
+    def add_line(self, text=""):
+        self._source += text + "\n"
+
+    def __str__(self):
+        return self._source
+
+
+prim = namedtuple("prim", ["name", "in_types", "return_types"])
+
+
+UInt = "UInt"
+SInt = "SInt"
+
+source = Source()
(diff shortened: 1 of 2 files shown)


leonardt:
I think that these wrappers cannot handle the bit manipulation required to match the widths to the coreir primitives because they do not know the widths of the inputs (i.e. we would need some notion of a width parameter `n` to know how many bits to select). The firrtl backend will need to properly emit the bit selection operations to have the outputs of these operations match the coreir primitives.

rdaly525:
I'm a little bit confused. Are you referring to bit slicing and concatenation? If so then yeah that might need to be a special case in the backend. 
Also for your ops do we need the product of Sint and Uint for the input types? I thought we were going to just create one firrtl module for each coreIR op where the all inputs/outputs to the firrtl module are always a UInt. I have no notion of signed and unsigned types, but I know you were trying to put those types into magma so it would probably be useful for you.

leonardt:
@rdaly525 Oh that makes sense, I interpreted the wording of your issue as wanting to wrap the firrtl primitive ops in modules but now I see you actually want the coreir ops implemented as a module.

leonardt:
I'll open another PR with an updated script

PR #140, opened 2017-08-11
Add script to generate FIRRTL modules for coreir primitives

leonardt:
Provides each coreir primitive as a FIRRTL module except `lshr`, `shl`, and `ashr` because I don't believe modules can have parameters.
The output widths of these modules will not match what coreir expects, so the FIRRTL backend will need to make sure to emit correct bit extraction, head, or tail operations after using one of these modules.


leonardt:
Also what is the different between `smod` vs `srem`?

rdaly525:
Thanks for doing this and sorry for the confusion. Don't throw away your original script though because it may be useful for firrtl to coreIR potentially?

There is a difference in calculation between signed remainder and signed modulo. I never remember exactly how they work though. You can ignore them for now.

leonardt:
@rdaly525 sounds good, I have the script saved on a local branch, I can push if it anyone is interested in using it.

PR #144, opened 2017-08-15
Stdlibchange

rdaly525:
This changes stdlib to coreir
Also changes in[2] to in0,in1.


rdaly525:
@leonardt, Can you fix the python issues due to changing in[2]->in0,in1? After that, feel free to merge and the hopefully your magma->coreir will work!

PR #153, opened 2017-08-17
Passes

rdaly525:
Passes are now either transform or analysis. Passes can specifiy dependencies (on analysis passes). There is also an api to get dependent analysis passes from the pass itself.  getAnalysis<T>()

Simplified PassManager to be very conservative.

Updated all the current passes and tests to use this new infrastructure

Added a couple of handwritten json files

Updated binary/coreir to allow for running passes and dynamically loading external passes (epasses/*)


rdaly525:
@leonardt, sorry this is a massive PR...

PR #154, opened 2017-08-17
Patch wire clock pass test

leonardt:
Minor patches to my simple wire clock example.
Test actually checks the presence of the new connections (as well as the lack of them initially).
Changes cast operations to use the `cast<>()` pattern. (Should this actually be a dynamic cast?)
Also, am I allowed to assume the top-level type of a moduleDef/Instance will always be a record?


leonardt:
I will merge #153 into this branch before we move this branch into `dev` as I see that #153 has made some changes

leonardt:
Merge complete, this should be ready to go.

PR #158, opened 2017-08-18
Verilog (but not verilog)

rdaly525:
Changes:
  Fixed up binary/coreir to have better UI
  Added necessary passes to create verilog (have not finished actual verilog pass)
    verify, strongverify, removebulkconnections,flattentypes
  On a whim created a custom coreirjson pass for "nice" json. 

Next PR will actually be the verilog!


rdaly525:
Fixed bug in LocalConnections to get Tests passing. Not sure why I am only seeing this now

PR #171, opened 2017-08-22
Ice40 example

leonardt:
Compiles through yosys, haven't had a chance to test it on the actual icestick yet.


rdaly525:
Awesome!!!! Thank you so much for helping out last minute

PR #192, opened 2017-09-04
Added simulator code to analysis passes

dillonhuff:
I've written a crude simulator pass that prints out the C code for clock level simulation of circuits. For now I put the unit tests in ./tests/unit/ and the simulator logic in ./src/coreir-passes/analysis.


dillonhuff:
Thanks.

Would you like the simulator tests to be in a subdirectory of ./test/unit? Eg ./test/unit/simulator for the unit tests themselves and ./test/unit/simulator/gencode for the auto generated files?

The 10000 line header is a header only unit testing framework that provides better printouts for test failures, allows test segmentation etc.

I added it because in my experience once you have even a moderate amount of unit tests written as isolated exes it gets pretty annoying to manage them, but its not a big deal and I can remove catch and rewrite the unit tests as individual files using asserts if you'd like.

rdaly525:
I have been meaning to revamp my testing framework, so this seems like it could be a good solution. Is there a guide you could point me to?

In terms of where the files should go, tests/simulator would probably be good.

dillonhuff:
Here is the [link to the catch github repo](https://github.com/philsquared/Catch)

I'll get started moving the files.

rdaly525:
One other thing that should make your life easier... I just created a doc: "doc/Casting.md" which explains how to do casting when using the coreir codebase. If you have worked with LLVM, it is exactly the same as that.

Also, the testing framework looks very useful. I am okay with adding it to coreir (and may try to convert some of my tests to use it). 

Another minor thing if it is not too bad, but it seems like you are gathering a lot of helper files in src/passes/analysis. Could you create a new subdirectory src/simulator and put all the files there? (but keep the files that actually do the passes)

And one last thing, what exactly do I need to do to try this out? I am excited to use it!

dillonhuff:
Thanks for adding documentation about casting. Its been a pain.

I've moved the simulator utilities files to ./src/simulator as you recommended.

To use the simulator first run the simulator pass on a coreir json file like you would any other pass:

```bash
bash-3.2$ ./bin/coreir -i ./examples/add4.json --load_passes ./lib/libcoreir.dylib -p simpass
```

Then the simulator generator will output 2 files, ```sim_add4.h``` and sim_add4.cpp```.

They look like so:

```cpp
#include <stdint.h>
#include <cstdio>
void simulate( uint16_t  self_in[ 4 ], uint16_t (*self_out_ptr) );
```
and sim_add4.cpp:
```cpp
#include <stdint.h>
#include <cstdio>
void simulate( uint16_t  self_in[ 4 ], uint16_t (*self_out_ptr) ) {

// Variable declarations

// Outputs
uint16_t self_out;

// Internal variables
uint16_t  i01_out;
uint16_t  i00_out;
uint16_t  i1_out;

// Simulation code
i01_out = self_in[2] + self_in[3];

i00_out = self_in[0] + self_in[1];

i1_out = i00_out + i01_out;

self_out = i1_out;

// Copy results to output parameters
(*self_out_ptr) = self_out;
}

```

The simulate function simulates one cycle of execution like verilator.

Also right now the simulator assumes the input file represents coreir that has already been flattened.

I'm not sure what kind of simulator API everyone would like so let me know if you have suggestions!

PR #196, opened 2017-09-10
Make install

leonardt:
Makefile changes:
* Renames existing `install` rule to be `build` (builds a copy of coreir in the cwd)
* Adds `install` rule that copies build and include files to `$(prefix)` which defaults to `/usr/local/` on Mac and `/usr` on Linux

Include file changes:
* renames top-level headers in `include/` to have a prefix `coreir-`. Should prevent name collisions with any other headers in `$(prefix)/include` (`passes.h`, `passmanager.h`, ... -> `coreir-passes.h`, `coreir-passmanager.h`, ...)

One issue:
* I'm a little worried about the glob (`*`) logic in the install rule. It makes it tricky to write a `make uninstall` rule where we remove any files that we add. One option is to explicitly list each file that's copied by `install`. This is cumbersome because there's a lot of headers and every time we add a new one we'll have to update the install rule. However if we have an explicit list of what files are installed, we'll be able to have an `uninstall` rule that removes them. Perhaps we can have a script that generates the install/uninstall commands with the files in the source directory? It seems like there should be an existing tool to do this (maybe `configure`). 

To test this on travis I migrated the python bindings to a separate repo (will submit a PR to remove them from this repo after we finish make install).
* Build passes: https://travis-ci.org/leonardt/pycoreir/builds/273890513
* The new bindings don't look in a specific place for the shared library, instead it expects it installed somewhere in the system path, see https://github.com/leonardt/pycoreir/blob/master/coreir/lib.py#L15 versus https://github.com/rdaly525/coreir/blob/master/bindings/python/coreir/lib.py#L17


rdaly525:
I think the uninstall issue could be solved by just having a base folder for coreir includes.
include/coreir/*

This would mean that uninstall would just be deleting the include/coreir

The current include directory structure is pretty terrible to begin with. This is a good time for me to migrate all of my header files into include/* 

I think I am going to update this branch to move all the header files into:

include/
  coreir-c/*.h
  coreir/
    ir/*.h
    passes/
      analysis/*.h
      transform/*.h
    libs/*.h
    
What do you think?

rdaly525:
Also... do you know how the coreir.h is actually including the "../src/ir/context.hpp"?
It seems very strange that it is finding that file.

leonardt:
That directory structure for the header files makes sense, no idea how that relative include is works. Now that I think about it, the python bindings only test if the dynamic library is installed correctly into the systems path. It doesn't check if the header files are working, so that relative include might not actually work.

I thought about updating the test suite to use the system installation rather than the build, but that seemed inconvenient for developing (where I would just want to run make build every time I made a change to the source).

Perhaps the best solution would be to have a set of "integration" tests that we run on travis that use the header files in `$(prefix)/include` and link the libraries in `$(prefix)/lib` (rather than appending the repo's build directories to `-I` and `-L` in the compile commands).

rdaly525:
I definitely agree with having tests run on the local build and having a couple integration tests.

rdaly525:
I am going to close this for now as I have all the changes migrated to #200

PR #202, opened 2017-09-13
Remove python bindings

leonardt:


leonardt:
@rdaly525 done!

leonardt:
Also you can now run `pip install coreir`!

PR #213, opened 2017-09-17
Argsimple

rdaly525:
Does the following:

Cleaned up the type select API.

Remove any notion of an "Any" type. I do not need this because I am doing static typechecking. For anything that needs no ports, we should use an empty Record.

Simplified the api for `Arg.get<T>`. Now I can simply pass in the type I am getting instead of the Arg's type.


rdaly525:
Yeah, you should get rid of any notion of "Any" in your bindings. Thanks!

Issue #214, opened 2017-09-17
Investigate using shared_ptr for Arg*


rdaly525:
@leonardt, if I did this, I think we could just keep a copy of the shared_ptr around in context for the C API? Or maybe there would be more problems than this.

PR #216, opened 2017-09-20
Changed Arg* to be a shared_ptr

rdaly525:
This allows me to not have to manage any memory for Args (with the exception of the c api)
This is the first part of a goal of mine to change Arg->Expr which will contain things other than just constants. This will likely eventually involve me creating my own intrusive pointer class (similar to Halide)

You can now call Const(5) in place of c->argInt(5)

Everything else should work the same including casting (That took a bit of work). 

A lot of the changes in the files are just changing Context.arg*() into Const()

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ ifeq ($(COREIRCONFIG),g++-4.9)
 CXX = g++-4.9
 endif
 
+CXX=clang++
+
 CFLAGS = -Wall -fPIC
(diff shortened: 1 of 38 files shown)


rdaly525:
@leonardt, can you verify that this works with pycoreir? I changed some of the c api implementations to be compatible with this.

rdaly525:
This will fix #214

leonardt:
@rdaly525 `pycoreir` test suite passes

PR #235, opened 2017-09-28
Memory example

dillonhuff:
the simulator now has support for ram


src/simulator/utils.hpp
```diff
@@ -80,6 +80,28 @@ namespace CoreIR {
 
     return genRefName == "reg";
   }
+
+  static inline bool isMemoryInstance(CoreIR::Wireable* fst) {
+    if (!isInstance(fst)) {
+      return false;
+    }
+
+    CoreIR::Instance* inst = toInstance(fst);
```
rdaly525: This can just be `CoreIR::Instance* inst = cast<Instance>(fst);`

PR #245, opened 2017-10-09
Configs

rdaly525:


Makefile
```diff
@@ -23,7 +23,8 @@ CXX = g++-4.9
 endif
 
 CFLAGS = -Wall -fPIC
-CXXFLAGS = -std=c++11  -Wall  -fPIC -Werror
+#CXXFLAGS = -std=c++11 -Wall -fPIC -Werror -ferror-limit=5
+CXXFLAGS = -std=c++11 -Wall -fPIC -Werror -Wno-error=sign-compare
```
leonardt: Should we fix the sign-compare issue instead of suppressing the warning? IIRC, C style mixed-sign comparisons can lead to difficult issues.

PR #262, opened 2017-10-20
fixed removeinst when there is one instance

jeffsetter:
Fixed removeinst when there is only one instance.
Also, added some warnings when using typegens.


rdaly525:
@leonardt, this change touches stuff with the C-API, can you verify that this works with pycoreir?

leonardt:
@rdaly525 works for me locally, passes on travis https://travis-ci.org/leonardt/pycoreir/builds/290989817?utm_source=email&utm_medium=notification

PR #269, opened 2017-10-24
Remove "Instantiable" keyword

rdaly525:
The goal of these changes was to remove the Instantiable Keyword as it was probably the most clunky part of the IR. 

Known changes
Major
- Removed Instantiable class. Technically I replaced it with GlobalValue which represents a constant in a particular namespace (Generator, Module, TypeGen, NamedTypes)
- Removed Instance.isGen, Instance.getGeneratorRef, Instance.getGenArgs, inst.getInstantiableRef
- Added Instance.getModuleRef, Module.isGenerated(), Module.getGenerator(), Module.getGenArgs
- Changed the C API to reflect these changes. 
- InstanceGraphPass now only loops over modules (and not generators)

Minor
- Modules now are forced to have a RecordType instead of a Type. 
- Separated GlobalValue, Generator, and Module into distinct files
- createInstanceMap is now a single pass that produces different maps for modules and generaotrs
- inline passthrough only works for "_.passthrough"

Most of the changes in this PR are replacing all Instantiables with either module or generator when appropriate.

@dillonhuff 
I also changed a bit of utils.hpp to show how to use the coreir-style casting
I changed your tests to run passes via the correct API. (You keep finding ways to do things in coreir that I do not expect, which is good for me to understand)


dillonhuff:
Looks good to me.

rdaly525:
@dillonhuff, do you mind taking a look at your simulator files again (in this PR) as a sanity check? I just had to update a chunk of your functions due to pushes in dev.

dillonhuff:
Looks good to me.

Thanks for modernizing utils too!

PR #270, opened 2017-10-24
Dev

rdaly525:
The changes here are mostly related to verilog/firrtl changes

Minor changes are `coreir.bit<op>` is now `corebit.<op>`
register with en,clr,rst are now in mantle
add with cin,cout is now in mantle.

@jeffsetter, @leonardt I think these changes should only affect halide and magma


jeffsetter:
Warnings are produced by interpret.cpp. Ideally there would be no comparison warnings.

PR #273, opened 2017-10-24
Typegen from python

leonardt:


leonardt:
@rdaly525 this is now passing, but there's one nagging issue that I think might be good to resolve when we merge this in. 

To include support for a python typegen, you need to set the `COREIR_INCLUDE_PYTHON_TYPEGEN=1`. This enables a preprocessor macro to include `Namespace::newTypeGenFromPython`. This also means you need to link `libpython`, which makes sense, but now it also means that any project using `libcoreir` needs to link `libpython` as well (as reflected by the fact that I needed to updated all the Makefiles to include the python link flags for them to pass). 

There's a two possible solutions that I think might ameliorate this issue. 

* Come up with a way (similar to our coreir-libs) to register the typegen with coreir without having to add a function to `Namespace`. This would enable coreir to be compiled without linking python and instead have a separate shared library `libcoreir-python` with the `typegen` defined, perhaps by created a new `python` namespace, that can be linked/loaded by anyone who needs the functionality (and consequently they will need to make sure they link libpython, but perhaps this is still better because at least they're explicitly requiring Python support by linking in the `libcoreir-python` shared library)
* Provide a `coreir-config` tool that provides a `--ldflags` commands that enables a user to easily add all the link flags necessary for their coreir installation (so, if they've compiled with Python support, this will link the required Python

leonardt:
Closing in favor of a less intrusive implementation via pycoreir https://github.com/leonardt/pycoreir/pull/16

PR #278, opened 2017-10-25
removed reset from dff for now

rdaly525:
I am going to remove the reset for now from the single bit FF in effort to get simulation working. Can you verify that this change does not break mantle/magma?


leonardt:
This will break Mantle's implementation of reset, is there a workaround? Can I add a mux on the input that selects the init value when reset is asserted?

PR #279, opened 2017-10-25
C Bindings for Interpreter

shacklettbp:


src/coreir-c/simulator-c.cpp
```diff
@@ -0,0 +1,116 @@
...
+    for (int i = 0; i < val_len; i++) {
+      bv.set(i, new_val[i]);
+    }
+    state->setValue(path, bv);
+  }
+
+  void CORESimStepMainClock(CORESimulatorState* cstate) {
+    SimulatorState* state = rcast<SimulatorState*>(cstate);
+    state->stepMainClock();
+  }
```
leonardt: It seems like wrapping method invocations on classes with no arguments or return values could be wrapped up in a simple script to generate it. We could create a simple script to manage these patterns and probably use it to generate the ctypes declaration for `pycoreir` as well. Maybe this is what SWIG does? Now that I'm thinking about it, I don't think it would be that hard to write a script to generate an implementation of the data-structure transfers from C++ to C to Python (e.g. methods that pass an UnorderedMap to key, value arrays in C to a python dictionary).

```
{{class}}.{{method}}
```

creates
```
void CORE{{class}}{{method}}(CORE{{class}}* arg) {
  rcast<{{class}}*>(arg)->{{method}};
}
```

PR #283, opened 2017-10-26
Interpreter cycle counter

dillonhuff:
At the request of @shacklettbp I have added a cycle counter. You can now read out total cycles executed by the clock. As part of this change the simulator now automatically detects the main clock and crashes with an error if the circuit has more than clock.

For example, if the circuit has one clock named "self.clk" you can get the number of cycles executed like so:

```cpp
ClockValue* clk = toClock(state.getValue("self.clk"));
int cycleCount = clk->getCycleCount();
```

### DIFF
--- a/src/binary/Makefile
+++ b/src/binary/Makefile
@@ -3,7 +3,7 @@
 HOME = ../..
 INCS = -I$(HOME)/include -I.
 LPATH = -L$(HOME)/lib
-LIBS =  -Wl,-rpath,$(HOME)/lib -lcoreir -ldl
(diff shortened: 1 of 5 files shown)


dillonhuff:
*more than one clock

rdaly525:
Instead of checking for the name self.clk, can you check for a signal of type corer.clk ?

dillonhuff:
A signal with type ```coreir.clk``` is what I check for in ```findMainClock()``` this is just an example where the name of the main clock happens to be ```self.clk```

rdaly525:
@dillonhuff Can you resolve the conflicts?

PR #287, opened 2017-10-27
Fixed the Arg inline issue

### LINKED ISSUE #286: Update inlining to correctly resolve modargs that have "arg"

rdaly525:
This should fix #286


dillonhuff:
Inlined linebuffer to verilog works now.

PR #289, opened 2017-10-30
Tried to fix linebuffer wraparound error when the depth of the buffer is not a power of 2

dillonhuff:
I've gotten the conv examples to run in the interpreter and compiled code simulator, but noticed that the linebuffer read and write indexes do not wrap around at the right place when the depth of the buffer is not a power of 2, so for example when the linebuffer in conv_3_1 has depth 10 it is indexed by a 4 bit address with a max value of 16, but the raddr, and waddr registers in linebuffermem just keep incrementing until they get to 15 and then reset since they are 4 bit numbers, causing the simulator to crash.

I have tried to fix this by updating the linebuffermem definition so that when depth is not a power of two the incremented value gets compared to the depth and set to 0 if the next value is equal to the depth.

I have also added LinebufferMem unit tests to /tests/simulator/interpreter.cpp for both the power of 2 and non-power of 2 cases.

@rdaly525 can you take a look and see if the changes to LinebuffereMem make sense?


rdaly525:
@dillonhuff, looks like this failed on travis. Could you investigate?

PR #294, opened 2017-11-02
Added symtable info for inlining

rdaly525:
Now after flattening, you should have a map from original signal to a valid wire in the flattened design. 

You can access this via something like the following:

```
map<string,json> symdata = topmodule->getMetaData()["symtable"].get<map<string,json>>();
SelectPath curpath = symdata["blah.blah"].get<SelectPath>();
```

Let me know if any issues arise


dillonhuff:
Does every module have metadata, or only the top module? For example in the unit tests this fails:

```cpp
      addCounter(c, g);

      uint pcWidth = 17;
      Type* counterTestType =
    	c->Record({
    	    {"en", c->BitIn()},
    	      {"clk", c->Named("coreir.clkIn")},
    		{"counterOut", c->Array(pcWidth, c->Bit())}});

      Module* counterTest = g->newModuleDecl("counterMod", counterTestType);
      ModuleDef* def = counterTest->newModuleDef();

      auto ct = def->addInstance("counter", "global.myCounter", {{"width", Const::make(c,pcWidth)}});

      def->connect("self.en", "counter.en");
      def->connect("self.clk", "counter.clk");
      def->connect("counter.out", "self.counterOut");

      counterTest->setDef(def);

      c->runPasses({"rungenerators", "flattentypes","flatten"});

      map<string,json> symdata =
        counterTest->getMetaData()["symtable"].get<map<string,json>>();

      cout << "symdata size = " << symdata.size() << endl;

      for (auto& symEnt : symdata) {
        SelectPath curpath = symdata[symEnt.first].get<SelectPath>();
        for (auto& p : curpath) {
          cout << p << "$";
        }

        cout << endl;
      }

```

With the error:

```bash
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
simtests is a Catch v1.3.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
Interpret simulator graphs
  Counter
-------------------------------------------------------------------------------
interpreter.cpp:71
...............................................................................

interpreter.cpp:240: FAILED:
due to unexpected exception with message:
  type must be object, but is null
```

PR #304, opened 2017-11-06
Dev

rdaly525:
Interpreter API updates
SMT added
Inline fixes

@leonardt, can you verify that this will not break anything in magma/mantle/pycoreir


leonardt:
@rdaly525 Everything working locally for pycoreir/magma/mantle

PR #320, opened 2017-11-12
Dev

rdaly525:
@leonardt if you have a chance could you verify magma, pycoreir, silica, etc?

I wonder if there is a way in the future to have a script which I can just run to verify everything whenever I want to PR -> master


leonardt:
pycoreir, magma, mantle, silica master branches all pass locally on my machine with the latest coreir dev build, looks like these changes were well scoped and didn't introduce any regressions!

It would be cool convenient to have Travis do this check. One approach would be to have Travis download all these repos and run their test suites when building a PR (if there's a travis environment variable that indicates it's a dev->master pull).

I know chisel has a set of regression tests that they run when doing a release.

@ucbjrl if you have a chance, could you give us brief description of how you handle regression testing for chisel releases?

ucbjrl:
We currently use Jenkins jobs to do a quick "does it break the build" suite of tests on every chisel3 pull request to the  master branch. The firrtl and dsptools repositories use Travis for similar checks. These tests typically complete in under two minutes.

We periodically poll the BIG4 repositories (chisel3, chisel-testers, firrtl, firrtl-interpreter) for any updates (typically merges) to their master branches and a change to any of these triggers a Jenkins job that checks out the master branches for all BIG4 repositories and any we know to be dependent on them (chisel-tutorial, chisel-template, rocket-chip, dsptools). Jenkins does an `sbt clean test publishLocal` in each of the cloned BIG4 workspaces, followed by a build and test in the dependent repo, using the freshly built BIG4 code.

The dependent repositories typically have `make` regression targets (or equivalent) and some mechanism to select the version of chisel3 (firrtl, firrtl-interpreter, chisel-testers) to build against. In the case of those dependent repositories which contain a chisel3 (firrtl, firrtl-interpreter, chisel-testers) submodule/subproject (currently rocket-chip), we overwrite their local checkout with the version of the master(s) under test. The full suite of regression tests takes approximately 45 minutes to complete.

Obviously, it's important to control the test environment to ensure we're actually testing what we think we're testing. We're moving to the Jenkins Pipeline architecture which allows us explicit control over the stashing of build artifacts for use in downstream pipeline stages, and the archiving of particular artifacts for either publishing (following a successful test completion) or post-mortem analysis (following a failure).

(This process should be written up and described in more detail.)

leonardt:
awesome thanks for the writeup, the design makes sense. Looks like we are starting to run into a similar problem so this will be a good reference for designing our testing infrastructure.

also, when things stabilize a bit more on our side it would be nice to integrate our software into your testing infrastructure (and vice versa)

PR #323, opened 2017-11-18
Add xnor to corebit

leonardt:


leonardt:
Xor -> not should work, I'll try that out

PR #335, opened 2017-12-11
Register inputs pass

dillonhuff:
Jeffs designs contain pure combinational paths from inputs to outputs that are problematic for both verilator and our simulator. This pass should eliminate the problem.


dillonhuff:
@rdaly525 Would you like more changes to this pass before merging? I would like to get it in ASAP to help with Jeffs use of the simulator

rdaly525:
Ill merge this for now. Ive opened an issue to fix the issue with registering all inputs for every module rather than just the top module.

PR #336, opened 2017-12-12
Add `simulator/` install rule to release makefile

leonardt:
This rule is missing in the release Makefile, unfortunately you'll need to do a new release to fix this issue. This breaks the pycoreir/magma/mantle/... master travis builds because they use the install rule in the release package.


leonardt:
There's likely a way to improve the install rules in both Makefiles to be more general rather than hardcoding each directory. Also, we should consider if there's an alternative to maintaining two separate Makefiles. We should at least enforce the practice that Makefile changes are only accepted if they update both.

PR #342, opened 2017-12-23
Grouping bitwise connections and adding an RTLIL library for Verilog to CoreIR

dillonhuff:
In order to get some larger designs for the simulator I've been working on a verilog frontend for coreir based on yosys.

The yosys frontend is in another repo on my github since it depends on github, but to make the translation simple I'm in the process of adding an RTLIL library for coreir as well as a pass that groups bitwise connections into single connections where possible (we discussed this pass with Lenny earlier because some Magma -> CoreIR translations also produce many bit by bit connections).

This PR contains enough of that library to get an older version of the CGRA loaded into coreir and to simplify its bit by bit connections.


rdaly525:
I did not realize you were working on this! This is great stuff!

For the group connections pass:
I have been meaning to write a bunch of tests which are basically just simple scripts which run bin/coreir and diff against a gold file. This seems like a good case for that type of test. My outputs should be deterministic, so diffing should work.

Something like:
>coreir -i test.json -p groupconnections -o test_out.json
>diff test_out.json gold.json


For the Yosys->CoreIR.
What do I need to do in order to use this? Do you know what the limitations are so far? (For example I am sure that we have no translation for inout wires).

dillonhuff:
Typo in my first post: "since it depends on github" should be "since it depends on yosys"

The verilog to CoreIR via yosys translation is located here:

[https://github.com/dillonhuff/VerilogToCoreIR](https://github.com/dillonhuff/VerilogToCoreIR)

I don't think yosys provides itself as a set of libraries, so for now the yosys to coreir translation works as a yosys pass. You can see some usage examples in the [makefile test section](https://github.com/dillonhuff/VerilogToCoreIR/blob/1c7e119a32200376070cc7dde8644715f949e3fe/Makefile#L1)

As far as limitations go there are a few things. For now I am just translating the yosys RTLIL into the coreir rtlil library, but I don't have translation of the coreir rtlil library into coreir primitives done yet. A few things like muxes and dffs are done, and I am working on more.

A few limitations so far:

* Hard to test translation for large designs b/c the flatten pass (needed to convert coreir rtlil modules into coreir primtives) is very slow for large designs
* RTLIL allows inputs to many binary operators (+, -, etc) to be different lengths, to translate this into coreir primitives we need signed / unsigned extend nodes. Does this already exist?
* No inout wires
* Yosys treats clocks as bits, there is no clkIn type, so we need primitives for casting a bit to a clock to translate yosys DFF / memory primitives into coreir registers / memory
* We have rtlil.dff / rtlil.adff generators in the coreir rtlil library, but we don't handle yosys memory constructs yet. Writing a generator in the rtlil library will be simple enough, but translating them to coreir memory primitives will be more complicated.
* Translating parametric modules is limited. We can translate a design that instantiates parametric verilog modules into coreir, but each instantiation of a parametric module produces a new coreir module [see here](https://github.com/dillonhuff/VerilogToCoreIR/blob/1c7e119a32200376070cc7dde8644715f949e3fe/to_coreir.cc#L330). Ideally we would be able to translate parametric verilog modules into generators.
* I haven't done any real correctness testing through the simulator since translation to coreir primitives isn't done yet.

PR #343, opened 2018-01-03
Extend

rdaly525:
Added the following coreir ops:

coreir.zext (zero extend)
coreir.sext (sign extend)

They both have genparams {"width_in":Int, "width_out":Int}


dillonhuff:
Looks good to me

PR #345, opened 2018-01-03
Added strip and wrap

rdaly525:
added coreir.strip and coreir.wrap.

both take in an genparam {"type":CoreIRType}

if you want to cast from a clk to a bit, pass in the clk type to strip and it will "strip" away the named type.
if you want to cast from a bit to a clk, pass in the clk type to wrap and it will "wrap" the bit into a clock type.
 you can get the clk type by calling `c->Named("coreir.clk")`

The final interface for strip will look something like:
{"in":Named(coreir.clkIn"), "out": Bit}

The final interface for wrap will look something like:
{"in":BitIn, "out": Named(coreir.clk)}

Currently this is very limited, bit it should (hopefully) work for clk and rst.


dillonhuff:
Looks good to me!

PR #351, opened 2018-01-11
added valid bit to linebuffer

jeffsetter:
Added recursive linebuffer with optional valid bit to commonlib.


rdaly525:
@jeffsetter + @dillonhuff, can you investigate why this is failing?

dillonhuff:
It looks like the issue is that this commit comments out the linebuffermem definition, [see here](https://github.com/rdaly525/coreir/blob/0a647fa0d44bef0d28974131854b32d742b24179/src/libs/commonlib.cpp#L502). linebuffermems are not a simulation primitive, so without a definition the simulator crashes on them since they are unsupported primitives

PR #356, opened 2018-01-16
Patch verilog

leonardt:
added a Verilog definition for corebit_const to corebitVerilog

externally linked modules were being emitted when they shouldn't. basically check and exit before adding them to `modList`.


leonardt:
Pushed a different change set, figured it out.

Basically we should add `vmod` to the modList if it has a definition but the module doesn't have a definition.
```
if (vmod->hasDef()) {
      ASSERT(!m->hasDef(),"NYI linking error"); //TODO figure out this better
 +    modList.push_back(vmod);
      return false;
}
```
but we shouldn't add it if it's externally defined

PR #365, opened 2018-01-29
Interpreter fix, even more rtlil primitives and more partial evaluation support

dillonhuff:
The title pretty much covers it. This fixes several interpreter errors that caused bit selects to be handled poorly and improves the definitions in the rtlil library


dillonhuff:
@rdaly525 this should fix David's simulation error

PR #369, opened 2018-01-31
Adding In Fully Parallel Map And Reduce

David-Durst:
This has fully parallel map and reduce implemented and tested. Note that while reduce over multiple time periods is in here, it is not ready for use yet. Also, the beginnings of some convolution stuff is here but disabled.


tests/simulator/Makefile
```diff
@@ -10,17 +10,17 @@
 #CXX = g++-4.9
 #endif
 #
-#CXXFLAGS = -std=c++11  -Wall  -fPIC -Werror
+#CXXFLAGS = -std=c++11  -Wall -g -fPIC -Werror
```
rdaly525: Can you get rid of this?

PR #370, opened 2018-02-01
Removed hasSel. should only use canSel

rdaly525:


rdaly525:
@dillonhuff , can you change your simulator stuff to use the "canSel" method instead of "hasSel"? I imagine this is actually the method that you want.

PR #375, opened 2018-02-06
Add valuemodule to c api

leonardt:
Pass modules as generator params through the c api!


src/coreir-c/values-c.cpp
```diff
@@ -67,6 +67,12 @@ extern "C" {
     return rcast<COREValue*>(ga);
   }
 
+  COREValue* COREValueModule(COREContext* cc, COREModule* mod) {
+    Context* c = rcast<Context*>(cc);
+    Value* ga = Const::make(c, (Module *)mod);
```
rdaly525: Is there a reason that this is not using an rcast?

leonardt: Too much C programming lately heh, good catch

PR #377, opened 2018-02-07
Add support for ConstModule to json pass

leonardt:
More support for modules as generator argument!


leonardt:
Actually, this might need some refinement, will reopen when ready.

PR #381, opened 2018-02-09
Revised equivalence checking

cristian-mattarei:
* fixed some bugs in the coreir smt translation
* revised the equivalence checking script 
* moved verification scripts into tools/formal_verification folder


bindings/python/examples/hello_module.py
```diff
@@ -0,0 +1,50 @@
+import coreir
```
rdaly525: We moved our pycoreir out of coreir repo a while ago. Can you remoe this file or add it to Lenny's pycoreir?

src/ir/inline.cpp
```diff
@@ -92,6 +92,25 @@ void PTTraverse(ModuleDef* def, Wireable* from, Wireable* to) {
 }
 }
 
+namespace {
```
rdaly525: What exactly is this change? Was there a bug that you fixed? or is this old?

PR #388, opened 2018-02-12
Jsonfix

rdaly525:
This changes the following Json syntax for a Record type from 

```
[Record, {
  "field1" : Type1,
  "field2" : Type2,
}]

```
To

```
[Record, [
  ["field1" , Type1],
  ["field2",  Type2],
]]

```

This change is needed to preserve record type ordering.


rdaly525:
No, I am not sure what you are referring to. Are you saying that the default value for a register is always 0, and that should not be the case? Anyways, this sounds like it should be in a separate issue

rdaly525:
@dillonhuff, let me know if you have any questions about regenerating the json files that your tests use. 

After your tests are fixed, I can merge this PR

PR #390, opened 2018-02-12
Checking all string names for valid syntax

rdaly525:


dillonhuff:
@rdaly525 When I run the tests on this branch I get a segfault in test/unit/constant_folding.cpp before even getting to the partial evaluation tests.

Is that what you are seeing when you run the tests?

rdaly525:
I can confirm that I am getting a segfault on the constant_folding test. This is a different bug than what I saw before.

rdaly525:
Reverted the issue and file a new bug to track #406 

@dillonhuff, you should see the original syntax bug now (in test partial_eval)

PR #391, opened 2018-02-12
Simplified everything to use toString()

rdaly525:
I no longer have a bunch of random methods to convert objects to strings. They are all called toString().

@leonardt Just for a small code review


rdaly525:
Fixes #391

PR #397, opened 2018-02-14
Changed warning for duplicate connection to an Error.

rdaly525:


dillonhuff:
@rdaly525 the connection error I see disappears when I comment out the tests in aetherlingSim.cpp.

@David-Durst Can you look into this?

David-Durst:
https://github.com/rdaly525/coreir/pull/405 should fix this.

PR #403, opened 2018-02-15
Staging

rdaly525:
This combines the following branches

jsonfix (fixing the record ordering)
bitinout (adds the bitinout)
additions to the BitVector class for proper serialization/deserialization


rdaly525:
@dillonhuff, This branch is where we need to update all the json files. I have done a handful of them already. Currently I am stuck on fifo_magma_json.json because it is 2000 lines of code. Do you know the magma file which this came from? It would be much better to just regenerate it.
@leonardt, Can you set up a quick branch of magma to use this coreir branch as a backend for compiling to coreir?

dillonhuff:
@rdaly525 I would just comment out that test. I've been meaning to remove it since it actually doesnt contain any assertions.

I originally added it when I was trying to test whether the interpreter could load json from magma

PR #415, opened 2018-02-21
added bsd license

rdaly525:


rdaly525:
I updated to use your suggestions.

leonardt:
Did you mean to include the LLVM license?

rdaly525:
Yeah, I am using some code from LLVM. I was under the impression from the license I had to reference it

leonardt:
ah makes sense, didn't realize that

PR #420, opened 2018-02-21
Binary

rdaly525:


src/binary/coreir.cpp
```diff
@@ -19,12 +21,59 @@
...
+bool fileExists(string name) {
+  std:: ifstream infile(name);
+  return infile.good();
+}
+
+void parseLib(string lib,string& libname,string& filename,vector<string>& lpaths) {
+  struct utsname unameData;
+  uname(&unameData);
+  string osname = unameData.sysname;
+  cout << "osname " << osname << endl;
```
leonardt: Do we want to leave this in stdout? I would probably comment/use a debug macro that I could disable for releases

src/binary/coreir.cpp
```diff
@@ -19,12 +21,59 @@
...
+    ASSERT(0,"Cannot support OS " + osname);
+  }
+  
+  vector<string> f1parse = splitString<vector<string>>(lib,'/');
+  string libfile = f1parse[f1parse.size()-1];
+  vector<string> f2parse = splitString<vector<string>>(libfile,'.');
+  
+  if (f1parse.size()==1 && f2parse.size()==1) { //Just passed in the name of the library
+    libname = lib;
+    string filecheck = "libcoreir-" + libname + "." + ext;
```
leonardt: Nice, seems convenient to not have to do the giant `libcoreir-` prefix

src/binary/coreir.cpp
```diff
@@ -19,12 +21,59 @@
...
+    for (auto path : lpaths) {
+      filename = path + "/" + filecheck;
+      if (fileExists(filename)) {
+        return;
+      }
+    }
+    ASSERT(0,"Cannot find lib " + lib);
```
leonardt: Could maybe print out the value of `lpaths` so the user knows where it looked (in case they forgot to add a path or added it to the wrong path)

src/binary/coreir.cpp
```diff
@@ -106,17 +155,23 @@ int main(int argc, char *argv[]) {
...
-      ASSERT(openLibHandles.count(file)==0,"Cannot REopen " + file);
-      void* libHandle = dlopen(file.c_str(),RTLD_LAZY);
-      ASSERT(libHandle,"Cannot open file: " + file);
+    vector<string> lpaths = {"/usr/local/lib","/usr/lib"};
+    if (auto lpath = std::getenv("LD_LIBRARY_PATH")) {
+      lpaths.push_back(lpath);
+    }
+    if (auto lpath = std::getenv("DYLD_LIBRARY_PATH")) {
+      lpaths.push_back(lpath);
+    }
```
leonardt: Might want to add these paths to the docs so users know where to install libraries

PR #422, opened 2018-02-23
Constify 

### LINKED ISSUE #406: Segfault when I constify ModuleDef.getInstances()
in moduledef.h:
when I change:

`    std::map<std::string,Instance*> getInstances(void) { return instances;}`

to

`    const std::map<std::string,Instance*>& getInstances(void) const { return instances;}`

I get segfaults.

rdaly525:
This fixes #406 

The problem was that the following APIs invalidate certain iterators
`ModuleDef.disconnect(), Wireable.disconnect(), Wireable.disconnectAll()` invalidates iterators from `Wireable.getConnectedWireables() `and `ModuleDef.getConnections()`

`ModuleDef.removeInstance()` invalidates iterators from `Wireable.getConnectedWireables(), ModuleDef.getConnections(), ModuleDef.getInstances()`

This was causing segfaults in the following pattern:

```
for (auto inst : def.getInstances() ) {
  if (shouldDelete(inst)) {
    def->removeInstance(inst)
  }
}
```

The fix I have been using is the following:

```
vector<Instance*> toDelete;
for (auto inst : def.getInstances() ) {
  if (shouldDelete(inst)) {
    toDelete.push_back(inst);
  }
}
for (auto inst : toDelete) {
  def->removeInstance(inst)
}
```


rdaly525:
I am still seeing test failures. @dillonhuff, Can you check to see if there are any times where you calling disconnect() or removeInstance() while iterating over one of the containers I described?

rdaly525:
It is also possible that this could affect the performance of inline

leonardt:
This seems like an issue with modifying the iterator while traversing it. Might be convenient to provide an API such as `def.removeInstances(instanceList)` or `def.removeInstances(instanceFilter)` where `instanceFilter` is a predicate function that chooses what instances to remove.

PR #445, opened 2018-02-27
Support for unknown values and high impedance values in bit vectors

dillonhuff:
Changed the bit vector class so that it can support unknown values 'x', and high impedance values, 'z'.

The logical operations do not yet support unknowns and high impedance values, but includes enough code to construct, serialize and de-serialize.

@rdaly525 x and z values can be used in both hex and binary construction, but for storage bit vectors will have to be stored as binary, since there are no hexadecimal characters for the values like: ```4'b1z11```

### DIFF
--- a/include/coreir/ir/dynamic_bit_vector.h
+++ b/include/coreir/ir/dynamic_bit_vector.h
@@ -225,7 +225,7 @@ namespace bsim {
     quad_value_bit_vector(const int N_) : N(N_) {
       //bits.resize(NUM_BYTES(N));
       bits.resize(N);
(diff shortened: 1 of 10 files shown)


rdaly525:
This is great! What are your semantics of high impedance values? I am a little unsure if they should be included... but we can discuss that in person.

dillonhuff:
I have not implemented the boolean ops for x and high impedance yet. I just added the machinery to store, construct, serialize and de-serialize them.

Also I just added corebit.term and coreir.term support in the PR

I dont have a strong feeling about whether it should even be possible to operate with high impedance values.

PR #455, opened 2018-03-02
Added jeffs ashr fix, added combinational memory tests, unknown bit now has definition

dillonhuff:


src/ir/valuecache.cpp
```diff
@@ -6,7 +6,33 @@ using namespace std;
...
+      if (lv.binary_value() > rv.binary_value()) {
+        return true;
+      }
+    } else if (lv.is_binary()) {
+      return true;
+    } else if (rv.is_binary()) {
+      return false;
+    } else {
+      if (!(lv.is_unknown() && rv.is_unknown())) {
+        cout << "lv = " << lv << endl;
```
rdaly525: Can you remove the cout statements?  Also does this take into account high impedance values?

dillonhuff:
Just removed the cout statements.

This PR does not handle high impedance values, but the CGRA high impedance values are removed by tristate pre-processing anyway.

PR #473, opened 2018-03-13
memory leaks

rdaly525:
Added new toString for recrodparams. 
Changed ordering of deletion in context to get rid of memory errors.
Fixed tpyecache for empty records. 
Fixed cullgraph memory leaks
deconstructor for verilog pass


src/ir/typecache.cpp
```diff
@@ -79,7 +86,7 @@ RecordType* TypeCache::getRecord(RecordParams params) {
     RecordType* r = new RecordType(c,params);
     
     //Inout just needs a single type
-    if (r->isInOut()) {
+    if (r->isInOut() || params.size()==0) {
```
leonardt: This seems like a strange edge case to be supporting, is there a case where you want to have an empty record type? Perhaps that should be an assertion or error?

rdaly525: It is actually pretty common. A lot of my testbenches have no inputs or outputs and therefore are just an empty record. They just instantiate the DUT.

leonardt: Ah, that makes sense

PR #477, opened 2018-03-15
Added coreir.neg to simulator

dillonhuff:


include/coreir/ir/dynamic_bit_vector.h
```diff
@@ -503,13 +501,6 @@ namespace bsim {
 
       return str;
     }
-
-    bool is_binary() {
-      for (int i = 0; i < bitLength(); i++) {
-        if (!get(i).is_binary()) return false;
-      }
-      return true;
-    }
```
leonardt: Is this function being removed intentionally? This will break pycoreir and the c-api. Is there another way we should be doing this?

include/coreir/ir/dynamic_bit_vector.h
```diff
@@ -309,7 +308,6 @@ namespace bsim {
       ind++;
 
       char format = str_raw[ind];
-      (void) format;
```
rdaly525: Without this, it will fail on Linux citing an unused variable warning.

leonardt: Strange, isn't it used in the assertion below? Also seems, to compile on travis

include/coreir/ir/dynamic_bit_vector.h
dillonhuff: I moved is_binary in the file. It is now [here](https://github.com/dillonhuff/coreir/blob/324ada898218155ef22e568632046e1d5882f529/include/coreir/ir/dynamic_bit_vector.h#L486)

dillonhuff: Also it is now ```const``` as it should be.

leonardt: got it

PR #480, opened 2018-03-15
adding posedge negedge support for reset on registers

rdaly525:
Also changed corebit.dff to corebit.reg
Also added corebit.regrst.


rdaly525:
@dillonhuff, I am getting some errors from the simulator. I tried to update it to use corebit.reg instead of corebit.dff. Do you mind taking a look?

dillonhuff:
For sure. Looking now

PR #497, opened 2018-03-21
Changed a bunch of connect and select functions to take arguments by const ref

dillonhuff:
the 16 x 16 CGRA now flattens to completion without the Killed: 9 error. Takes ~5 minutes almost all of which is flattening


dillonhuff:
For reference I just timed yosys doing same flattening task and then outputting to verilog. It took 14 seconds, so yosys is around 20x faster

PR #513, opened 2018-04-06
Add support for initializing and getting BitVector hex strings through C API

shacklettbp:


src/coreir-c/values-c.cpp
```diff
@@ -52,6 +52,13 @@ extern "C" {
     *val = bv.to_type<uint64_t>();
   }
 
+  void COREValueBitVectorGetString(COREValue* a, char *dst) {
+    Value *value = rcast<Value*>(a);
+    BitVector bv = value->get<BitVector>();
+    string str = bv.hex_string();
+    memcpy(dst, str.c_str(), str.size());
+  }
```
leonardt: Looks good, just to clarify, the caller will need to allocate `dst` and I'm guessing that if it has access to the BitVector object, it can compute the required size of the buffer (specifically `str.size()`) ahead of time.

shacklettbp: Yeah that's correct, not super elegant but I didn't want to have to free the string manually in python.

PR #514, opened 2018-04-10
Changing connection comparison to just use pointer comparison

dillonhuff:
@rdaly525 I've been playing with ways to improve performance. The connection comparison function was a huge performance drain because it was iterating over select paths and comparing strings.

Since it looked like all the connection comparison function needed to do was provide some arbitrary order for pairs of pointers I changed it to just compare the pointer values.

This resulted in about a 40% improvement in flattening performance for the 16 x 16 CGRA. Is this change acceptable or does connection comparison need to compare select strings for some other reason?


rdaly525:
My one concern is the ability to serialize coreir. I think this should be fine as long as we do select path comparison for the coreir2json

rdaly525:
Yes, that is correct. It would also be great if you did the same comparison in the string toString(Connection) function (in src/ir/common.cpp)

dillonhuff:
@rdaly525 just added those changes

PR #529, opened 2018-04-26
Tutfix

rdaly525:
This pull request fixes the non-pass tutorials. I have also added a good amount more description for the IR here. 

Also some minor bug fixes in coreir.


rdaly525:
@David-Durst, Can you fix your test in this branch?

David-Durst:
See https://github.com/rdaly525/coreir/pull/530, which fails on another test that I do not believe is mine.

PR #531, opened 2018-05-03
Bunch of fixes in order to get linking to work

rdaly525:
Changed the Json Spec/fileReader/fileWriter/coreirjson. Removed namedtypegens, but added typegens. 
  Generated Modules are now inside generators.

Updated some json files in the repo to reflect these changes.

Added a ‘runPassesOnAll’ which will run passes on all namespaces.

Added a simpler saveToFile.

Updated ASSERT to require semicolon 

Added method to remove generated module

Reorganized namespace methods

Typegen:
  Added a ‘sparse’ typegen.
  Changed typegen to use a static make method.
  Added automatic caching of generated types

Added an unresolved symbols pass. which will delete everything except unresolved symbols

Add a -a option in binary to run on all namespaces

Reorganized coreir binary , and now allows an unspecified ‘top’


rdaly525:
@leonardt, sorry it is a bunch of changes. I will try to do smaller push requests in the future.

doc/JsonSpec.md
```diff
@@ -65,13 +68,14 @@ Wireable = "<instname>,<a>,<b>,..."
 
 
 //The following is my Value IR. 
-//This contains a small IR representing constants and Referneces to generator/module args. This will be expanded
+//This contains a small IR representing constants and Referneces to generator/module args. This can be expanded
```
leonardt: Typo `Referneces` should be `References`

include/coreir/ir/context.h
```diff
@@ -157,6 +160,7 @@ bool loadFromFile(Context* c, std::string filename,Module** top=nullptr);
 //Save namespace to a file with optional "top" module
 bool saveToFile(Namespace* ns, std::string filename,Module* top=nullptr); //This will go away
 bool saveToFilePretty(Namespace* ns, std::string filename,Module* top=nullptr);
+bool saveToFile(Context* c, std::string filename); //This will go away
```
leonardt: What does `This will go away` imply?

include/coreir/ir/typegen.h
```diff
@@ -3,36 +3,63 @@
 
 #include "fwd_declare.h"
 #include "globalvalue.h"
+#include "common.h" //This should really be moved to fwd_declare
```
leonardt: Then move it!

include/coreir/ir/typegen.h
```diff
@@ -3,36 +3,63 @@
...
+    TypeGenFromFun(Namespace* ns, std::string name, Params params, TypeGenFun fun, bool flipped) : TypeGen(ns,name,params,flipped), fun(fun) {}
+    Type* createType(Values genargs) override {
+      return fun(this->getContext(),genargs);
+    }
   public :
-    TypeGenFromFun(Namespace* ns, std::string name, Params params, TypeGenFun fun, bool flipped=false) : TypeGen(ns,name,params,flipped), fun(fun) {}
-    Type* createType(Context* c, Values genargs) override {
-      return fun(c,genargs);
+    bool hasType(Values genargs) override {
+      return true; //Really should check if params are val TODO
```
leonardt: I'm not sure what this new `hasType` is supposed to do, perhaps checks if the parameters a valid set of arguments that will result in a type?

src/ir/fileReader.cpp
```diff
@@ -245,6 +300,7 @@ bool loadFromFile(Context* c, string filename,Module** top) {
 
     //If top exists return it
     if (top && j.count("top")) {
+      c->getNamespace("global")->print();
```
leonardt: `loadFromFile` is now over 200 lines long, you should consider decomposing it into smaller functions to keep things better organized. It's hard for me to understand these changes because they touch so many different parts of the process. If each chunk was contained in a function with a well scoped and clearly defined purpose, it would be easier to understand what each change was responsible for.

src/ir/fileReader.cpp
```diff
@@ -160,46 +197,64 @@ bool loadFromFile(Context* c, string filename,Module** top) {
...
           if (jgen.count("defaultgenargs")) {
             g->addDefaultGenArgs(json2Values(c,jgen.at("defaultgenargs")));
           }
           if (jgen.count("metadata")) {
             g->setMetaData(jgen["metadata"]);
           }
+          for (auto vjpair : genmodvalues) {
+            cout << "in loop:" << endl;
+            Module* m = g->getModule(vjpair.first);
+            m->print();
```
leonardt: Do we need these cout/print functions in stdout by default? Should consider adding some logging infrastructure so you can keep `stdout` clean for applications that don't require this information.

src/ir/namespace.cpp
```diff
@@ -58,26 +59,23 @@ NamedType* Namespace::newNamedType(string name, string nameFlip, Type* raw) {
...
+// 
+//  //Add name to typeGenNameMap
+//  typeGenNameMap[name] = nameFlip;
+//  typeGenNameMap[nameFlip] = name;
+//
+//  //Create the TypeGens
+//  TypeGenFromFun::make(this,name,genparams,fun,false);
+//  TypeGenFromFun::make(this,nameFlip,genparams,fun,true);
+//  
+//}
```
leonardt: Just remove it now? You can always go back to a commit to find it (if you give your commit an easily identifiable message)

leonardt:
Functionality changes seems reasonable but it's challenging for me to understand whether it will work, but I think i get what new features have been added/changed (as per my summary above). Also added a couple inline style comments.

PR #538, opened 2018-05-17
Lb fixes

jeffsetter:
Fixes linebuffer implementation that was found while simulating against Halide. Also includes some commits in dev.


jeffsetter:
Since this was instead merged into dev, can we merge dev into master? These changes are necessary for my Halide changes to merge into master.

PR #539, opened 2018-05-18
Modularized loading libraries

rdaly525:
Added class for dynamic library loading. Added wrapper class for coreir library loading. simplified coreir binary to use this class for loading libraries

This temporarily removes the capability of dynamically loading passes.


rdaly525:
With latest commits, the dynamically loading passes should be working again.

PR #553, opened 2018-06-18
Jsontype

rdaly525:
Added a json type to Values.

### DIFF
--- a/include/coreir/ir/value.h
+++ b/include/coreir/ir/value.h
@@ -25,6 +25,7 @@ U2V_SPECIALIZE(BitVector,ConstBitVector);
 U2V_SPECIALIZE(std::string,ConstString);
 U2V_SPECIALIZE(CoreIR::Type*,ConstCoreIRType);
 U2V_SPECIALIZE(CoreIR::Module*,ConstModule);
+U2V_SPECIALIZE(Json,ConstJson);
 
 #undef U2V_SPECIALIZE
 
@@ -37,9 +38,9 @@ class Value {
       VK_ConstString=3,
       VK_ConstCoreIRType=4,
       VK_ConstModule=5,
-      VK_ConstEnd=6,
-      VK_Arg=7,
-      VK_CPPLambda=8
+      VK_ConstJson=6,
+      VK_ConstEnd=7,
+      VK_Arg=8,
     };
   private :
     ValueKind kind;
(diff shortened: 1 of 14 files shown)


rdaly525:
@leonardt, could your review this when you have a chance?

include/coreir/ir/value.h
```diff
@@ -99,22 +101,6 @@ class Arg : public Value {
...
-//    static std::shared_ptr<CPPLambda> make(LambdaType lambda) {
-//      return std::make_shared<CPPLambda>(lambda);
-//    } 
-//  private :
-//    LambdaType lambda;
-//  public :
-//    bool operator==(const Value& r) const {
-//      return false;
-//    }
-//};
```
leonardt: Should this be removed? You removed `VK_CPPLambda` from the enum list, but you kept this around. Should either keep both commented out for consistency (so you don't forget it if you're going to add it back in), or just removed both of them.

include/coreir/ir/value.h
```diff
@@ -206,9 +194,6 @@ class TemplatedConst : public Const {
     bool operator<(const Value& r) const override;
 
     static bool classof(const Value* v) {return v->getKind()==Underlying2Kind<T>::kind;}
-    //static std::shared_ptr<TemplatedConst<T>> make(ValueType* type, T value) {
-    //  return std::make_shared<TemplatedConst<T>>(type,value);
-    //}
```
leonardt: Not sure why the two above methods are removed.

include/coreir/ir/valuecache.h
```diff
@@ -10,16 +10,22 @@ class BitVectorComp {
     bool operator() (const BitVector& l, const BitVector& r) const;
 };
 
+//class JsonComp {
+//  public:
+//    bool operator() (const Json& l, const Json& r) const;
+//};
```
leonardt: Do you still need this?

src/ir/value.cpp
```diff
@@ -103,30 +103,13 @@ Const* Const_impl<Module*>(Context* c,Module* val) {
...
-//
-//template<>
-//string TemplatedConst<std::string>::toString() const {
-//  return vtype->toString() + "(" + value + ")";
-//}
-//
-//template<>
-//string TemplatedConst<Type*>::toString() const {
-//  return vtype->toString() + "(" + value->toString() + ")";
-//}
```
leonardt: All these toString methods are being removed?

rdaly525:
I do not have plans on using any other json library other than nholmann's.

include/coreir/ir/value.h
rdaly525: This was never in the code base. It was always commented. Probably should have removed it in a different PR

src/ir/value.cpp
rdaly525: Similar to the other comment, these were being unused. there was already a toString method defined for this class.

PR #554, opened 2018-06-18
Laxparams

rdaly525:
This creates a ValueType called "Any". This allows a Value of any type to be passed as generator arguments.

This also has some casting operations to facilitate with the Any type

(Do this after PR for jsontypes)


rdaly525:
@leonardt, when you get a chance, could you also review this PR

PR #560, opened 2018-06-19
Revive firrtl generation pass

edwardcwang:
@rdaly525

### DIFF
--- a/include/coreir/definitions/coreFirrtl.hpp
+++ b/include/coreir/definitions/coreFirrtl.hpp
@@ -1,7 +1,9 @@
-
-
 using namespace CoreIR;
 using namespace std;
+
+// The coreir namespace is for multi-bit primitives.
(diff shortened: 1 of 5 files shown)


edwardcwang:
Notebook: https://gist.github.com/edwardcwang/d21f4be167f2c186c27426c7dd5b43fd

edwardcwang:
@leonardt @rdaly525 One important observation (which might be actually be a bug in Magma) - in the magmathon jupyter notebook, `m.compile('build/add', main, output="firrtl")` doesn't seem to work (generates the old invalid firrtl).

I have to set output to "coreir" and manually run `coreir -i build/add.json -o build/add.fir` to get the kind of valid firrtl that this PR generates.

I also tried restarting the Jupyter notebook to no avail.

edwardcwang:
Resolved: apparently the magma firrtl emitter is different from the coreir firrtl emitter.

PR #578, opened 2018-07-16
Rename rowbuf `and` instance

leonardt:
Using an instance with the same name as an input port causes issues in generating verilog, see
https://github.com/YosysHQ/yosys/issues/582#issuecomment-405396063 for an example (there's an instance named `valid` for the rowbuffer and a port named `valid`).

This change is purely internal to the rowbuffer so shouldn't require any changes externally.

### DIFF
--- a/src/ir/headers/memories.hpp
+++ b/src/ir/headers/memories.hpp
@@ -9,9 +9,9 @@ bool isPowerOfTwo(const uint n) {
 }
 
 Namespace* CoreIRLoadHeader_memory(Context* c) {
-  
+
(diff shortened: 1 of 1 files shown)


leonardt:
Apologies for the whitespace included in the diff, there's a setting under files changed "Diff settings" to exclude whitespace.

leonardt:
Use this link for a diff without whitespace changes: https://github.com/rdaly525/coreir/pull/578/files?utf8=%E2%9C%93&diff=unified&w=1

PR #581, opened 2018-07-20
Passargs

rdaly525:
This PR adds the feature of being able to have arguments passed into the passes similar to yosys. This should be very helpful for writing passes.

Changes:
Added an optional initialize function to passes which takes in argc/argv so that they can be parsed like normal command line args.

coreir binary is slightly different now.
old: coreir -p rungenerators,flatten,verifyconnectivity-onlyinputs
new: coreir -p "rungenerators; flatten; verifyconnectivity --onlyinputs"

(passes are separated by a semicolon)

Pass Dependencies now depend on the combination of the passes *and* their arguments.

I currently have one example of a pass using args (verifyconnectivity --onlyinputs --noclkrst)
(This is why there are 26 files changed)

In the future, I should possibly rethink the way that passes are allocated. I should ideally create pass objects based off their arguments so that passes of the same kind with different arguments are actually real separate passes. Currently I conservatively invalidate all passes independent of their args during transforms.


src/ir/passmanager.cpp
```diff
@@ -118,10 +115,21 @@ bool PassManager::runInstanceGraphPass(Pass* pass) {
...
+  int argc = pArgs.size();
+  char** argv = new char*[argc];
+  for (int i=0; i<argc; ++i) {
+    argv[i] = &(pArgs[i])[0];
+  }
+  cout << "Numargs=" << argc << endl;
+  if (argc > 1) {
+    p->initialize(argc,argv);
+  }
+  delete[] argv;
```
leonardt: Good delete! 👍

src/ir/passmanager.cpp
```diff
@@ -154,59 +162,68 @@ bool PassManager::runPass(Pass* p) {
...
     //Actually run the passes now
     while (!work.empty()) {
-      string pname = work.top(); work.pop();
-      bool anal = analysisPasses.count(pname) > 0;
+      string passString = work.top(); work.pop();
+      vector<string> pArgs = splitStringByWhitespace(passString);
+      string pname = pArgs[0];
+      
       Pass* p = passMap[pname];
+      bool anal = p->isAnalysis;
```
leonardt: Might be a better variable name than this :p

PR #583, opened 2018-07-22
Added hellocounter into tests

rdaly525:


tutorial/hellocounter/run
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x #echo on
+echo "$1"
+
+#function mytest {
+#  "$@"
+#  local status = $?
```
leonardt: Do we need these comments?

tutorial/hellocounter/run
```diff
@@ -0,0 +1,13 @@
...
+set -e
```
leonardt: Pushed a commit to combine `set -ex` into one line

PR #591, opened 2018-08-04
Prevent Nested Clocks In Aetherling's Map

David-Durst:
As https://github.com/phanrahan/magma/issues/238 notes, the CoreIR wire-clocks pass fails when it has to deal with ports that are arrays of clocks.

This commit prevents map from lifting clock ports. Therefore, the wireclocks pass will wire up the clocks for each of the nested instances that have a clock port and doesn't have to deal with a top layer, array of clock ports.


David-Durst:
Closing as #592 is a more general solution to the problem.

PR #598, opened 2018-08-14
Add ModuleDefCanSelect to c-api

leonardt:
Enables a pycoreir check that raises an exception on selection errors instead of triggering a segfault in coreir.

See https://github.com/leonardt/pycoreir/commit/409c5e05c2238888348af27f8feaf2215985b224


leonardt:
Bump, this is one is a nice quality of life feature for pycoreir that would be nice to get in. Pretty common segfault case (selecting an invalid port).

PR #601, opened 2018-08-20
Add c api methods for metadata

leonardt:
First pass support for passing simple string key, value pairs for metadata. This allows magma to pass the 'filename' and 'lineno' fields to coreir.

Basically
```
rcast<Wireable*>(wireable)->getMetaData()[key] = value;
```

If we can make the `Connections` type (currently a typedef to pair) a `MetaData`, we can also add support for filename/lineno passing for connections.

Actually, could we change `COREWireableAddMetaDataStr` to `COREMetaDataAddMetaDataStr`? (I think this should just work for any class that subclasses from MetaData).

### DIFF
--- a/include/coreir-c/coreir.h
+++ b/include/coreir-c/coreir.h
@@ -19,7 +19,7 @@ void* CORENewMap(COREContext* c, void* keys, void* values, uint len, COREMapKind
 extern COREContext* CORENewContext();
 extern void COREDeleteContext(COREContext*);
 extern COREType* COREContextNamedType(COREContext* context, const char* namespace_, const char* type_name);
-  COREType* COREContextFlip(COREContext* context, COREType* type);
+extern COREType* COREContextFlip(COREContext* context, COREType* type);
 
 extern COREValueType* COREContextBool(COREContext* context);
 extern COREValueType* COREContextInt(COREContext* context);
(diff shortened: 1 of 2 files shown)


leonardt:
With this, this magma file
```python
import os
import magma as m
from magma.testing import check_files_equal
import pytest


@pytest.mark.parametrize("target,suffix",
                         [("verilog", "v"), ("coreir", "json")])
def test_simple_def(target, suffix):
    m.set_codegen_debug_info(True)
    And2 = m.DeclareCircuit('And2', "I0", m.In(m.Bit), "I1", m.In(m.Bit),
                            "O", m.Out(m.Bit))

    main = m.DefineCircuit("main", "I", m.In(m.Bits(2)), "O", m.Out(m.Bit))

    and2 = And2()

    m.wire(main.I[0], and2.I0)
    m.wire(main.I[1], and2.I1)
    m.wire(and2.O, main.O)

    m.EndCircuit()

    m.compile("build/test_simple_def", main, output=target)
    m.set_codegen_debug_info(False)
    assert check_files_equal(__file__, f"build/test_simple_def.{suffix}",
                             f"gold/test_simple_def.{suffix}")
```
produces the json
```json
{"top":"global.main",
"namespaces":{
  "global":{
    "modules":{
      "And2":{
        "type":["Record",[
          ["I0","BitIn"],
          ["I1","BitIn"],
          ["O","Bit"]
        ]]
      },
      "main":{
        "type":["Record",[
          ["I",["Array",2,"BitIn"]],
          ["O","Bit"]
        ]],
        "instances":{
          "inst0":{
            "modref":"global.And2",
            "metadata":{"filename":"tests/test_circuit/test_define.py","lineno":"16"}
          }
        },
        "connections":[
          ["self.I.0","inst0.I0"],
          ["self.I.1","inst0.I1"],
          ["self.O","inst0.O"]
        ]
      }
    }
  }
}
}
```

PR #602, opened 2018-08-21
Connmeta

rdaly525:
Adds api for adding metadata to connections

Unfortunately it was going to be an immense amount of changes to have a connections class inheret from MetaData, so I have a separate API from ModuleDef to add metadata to connections.

(I am still working on a test here, but I figure this can get you started)


leonardt:
I see {get,has}MetaData, I think I need a setMetaData as well

PR #606, opened 2018-08-22
Dev

rdaly525:


leonardt:
@rdaly525 anything blocking this? I believe we were holding off because of the magma hackathon. Now that its over, I think we can consider merging this in? We can run a flow build to make sure that isn't broken.

PR #607, opened 2018-08-27
CoreIR Changes/Fixes for Integration with CGRA-ME

JKHHai:
Makes changes to the CoreIR library so that it works with CGRA-ME.

Main changes:
- `RemoveBulkConnections`: Fixed a bug where Inouts were not recognized as bit connections
- `addDefaultGenArgs`: Fixed a bug where the default parameters were not being propagated to the Verilog writer.

To do:
- Enable Verilog generation of Inout connections using CoreIR

### DIFF
--- a/src/ir/moduledef_validate.cpp
+++ b/src/ir/moduledef_validate.cpp
@@ -20,15 +20,41 @@ bool ModuleDef::checkTypes(Wireable* a, Wireable* b) {
   //TODO This might not be valid if:
   //  2 outputs are connected to the same input
 
-  //cout << "Got types" << endl;
-  
   if (ta == c->Flip(tb) ) {
-    //cout << "ta flipped" << endl;
     return false;
   }
 
-  //cout << "Flipped types" << endl;
-  
+
+// Potential fix for Inout Connections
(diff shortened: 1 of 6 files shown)


src/ir/moduledef_validate.cpp
```diff
@@ -20,15 +20,41 @@ bool ModuleDef::checkTypes(Wireable* a, Wireable* b) {
...
+//        cout << "1 BitInOut Array, 1 " << tb->toString() << endl; 
+//        return false;
+//      }
+//      if ((tb->elemType->getKind() == tb->TypeKind::TK_BitInOut) && ((ta->elemType->getKind() == ta->TypeKind::TK_BitIn) | (ta->elemType->getKind() == ta->TypeKind::TK_Bit)))
+//      {
+//        cout << "2 BitInOut Array, 1 " << ta->toString() << endl; 
+//        return false;
+//      }
+//    }
+//  }
```
leonardt: Can you submit the code in the comments as a separate PR? Or you could include it in this PR, but it would be good to include a test case that shows what this fixes.

Issue #609, opened 2018-08-27
Add option to split into multiple verilog files


rsetaluri:
See: https://github.com/phanrahan/magma/issues/254

rsetaluri:
Solved by #612

PR #612, opened 2018-08-30
Split verilog files

rsetaluri:
See #609.

### DIFF
--- a/src/binary/coreir.cpp
+++ b/src/binary/coreir.cpp
@@ -1,6 +1,7 @@
 #include "coreir.h"
 #include "coreir/tools/cxxopts.h"
 #include <fstream>
+#include <memory>
 #include "passlib.h"
 
 
@@ -41,6 +42,7 @@ int main(int argc, char *argv[]) {
     ("t,top","top: <namespace>.<modulename>",cxxopts::value<std::string>())
     ("a,all","run on all namespaces")
     ("z,inline","inlines verilog primitives")
+    ("s,split","splits output files by name (expects '-o <path>/*.<ext>')")
     ;
   
   //Do the parsing of the arguments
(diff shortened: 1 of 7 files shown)


src/binary/coreir.cpp
```diff
@@ -171,8 +188,12 @@ int main(int argc, char *argv[]) {
     cout << "Running vpasses" << endl;
 
     auto vpass = static_cast<Passes::Verilog*>(c->getPassManager()->getAnalysisPass("verilog"));
-    
-    vpass->writeToStream(*sout);
+
+    if (split_files) {
```
rsetaluri: @rdaly525 right now this flag is only considered in verilog mode. Other modes will ignore it altogether. What would you like to do here. I can add the stub method `writeToFiles()` for each of the other passes (where the implementation is to raise a not-implemented error. Otherwise we can leave it as is and document that this only works for verilog mode.

src/binary/coreir.cpp
```diff
@@ -119,23 +122,37 @@ int main(int argc, char *argv[]) {
...
+             || outExt == "smt2"
+             || outExt == "smv"
+             || outExt == "v", "Cannot support out extention: " + outExt);
+      if (!split_files) {
+        std::unique_ptr<std::ofstream> fout(new std::ofstream(outfile));
+        ASSERT(fout->is_open(),"Cannot open file: " + outfile);
+        sout = fout.release();
+      }
+    }
+    if (split_files) {
```
rdaly525: Can you add an assert here that the output file extension is verilog? and print a "NYI" otherwise.

rsetaluri: Done

rsetaluri:
@rdaly525 not seeing where this changes the default behavior. I believe if `split_files` is false, then the codepath is the same as before (modulo some refactoring, but functionally still the same). Can you point out where the difference is?

PR #620, opened 2018-09-20
Newapp fix

jeffsetter:
Applies some fixes in commonlib and memories for harris to work with the Halide


jeffsetter:
We have a passing cgraflow with this branch:
https://travis-ci.org/StanfordAHA/CGRAFlow/builds/431619563

src/ir/headers/memories.hpp
```diff
@@ -28,6 +28,7 @@ Namespace* CoreIRLoadHeader_memory(Context* c) {
   });
 
   //Note this is a linebuffer MEMORY (a single row) and not a full linebuffer.
+  //memory->newGeneratorDecl("rowbuffer",memory->getTypeGen("rowbufferType"),MemGenParams);
```
leonardt: Do we need this comment?

src/ir/headers/memories.hpp
```diff
@@ -282,20 +283,63 @@ Namespace* CoreIRLoadHeader_memory(Context* c) {
...
+    def->connect("waddr0.out","mem.waddr");
+    def->connect("wdata0.out.0","mem.wen");
+    def->connect("mem.rdata","readreg.in");
+    def->connect("self.rdata","readreg.out");
+    def->connect("self.raddr","raddr_slice.in");
+    def->connect("raddr_slice.out","mem.raddr");
+    def->connect("self.ren","readreg.en");
+  });
+
+
```
leonardt: Adding this new generator shouldn't affect any existing code

leonardt: Can you add a comment explaining how this rom (rom2) is different than the normal rom? Also, is there a better (more descriptive) name than just `rom2`?

leonardt: From what I understand, this module sets waddr/wdata/wen to 0 (since it's a rom), then passes the rdata through a register (adds a single cycle delay). It also seems to slice the `raddr` (so it only selects the lower bits based on the depth). Why can't the interface to this ROM have the `raddr` width equal to clog2(depth)? This would remove the need for the slice module.

leonardt: Maybe `rom_delay_1` as a name (we should figure out a better way to handle parametrized delays, but this is okay for now to unblock you).

src/libs/commonlib.cpp
```diff
@@ -307,6 +316,78 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
+      for (uint i=0; i<out_wires.size(); ++i) {
+        std::string const_name = "const_" + std::to_string(i);
+        Values const_args = {{"width", Const::make(c,bitwidth)}};
+
+        Values const_configargs = {{"value", Const::make(c,BitVector(bitwidth,value))}};
+        Wireable* const_inst = def->addInstance(const_name, "coreir.const", const_args, const_configargs);
+        def->connect(const_inst->sel("out"), out_wires[i]);
+      }
+
+    });
```
leonardt: Can we get documentation for this generator?

leonardt: It looks like a convenience generator for creating an array of constants?

src/libs/commonlib.cpp
```diff
@@ -802,33 +883,96 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
+               ", dimension length must be divisible, because we can't swizzle data");
+
+        if (img_dimx - out_dimx < 3 && (img_dimx != out_dimx)) {
+          std::cout << "Image dimension " << dim << "  is " << img_dimx
+                    << " and output stencil size is " << out_dimx
+                    << ", which means the linebuffer mem is going to be very small"
+                    << std::endl;
+      
+        }
+      }
```
leonardt: Can you separate this for loop into a helper function? (such as `check_dimensions_are_correct`), that would help make this code more organized.

src/libs/commonlib.cpp
```diff
@@ -839,9 +983,115 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
+            string source = out_pair.first;
+            string sink = out_pair.second;
+            uint iflip = inverted_index(outx, inx, i);
+            out_temp.push_back({source + "." + to_string(i),
+                    sink + "." + to_string(iflip)});
+            
+          }
+        }
+        out_pairs = out_temp;
+      }
```
leonardt: This loop is also another good candidate for a helper function (e.g. `make_out_pairs`), also a comment explaining what this is doing would be good.

src/libs/commonlib.cpp
```diff
@@ -1034,63 +1232,48 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
+        //   size_lbmems = prod((img[x] - (out[x]-in[x])) / in[x]) 
+        //      except for x==1, img0 / in0
         uint size_lbmems = 1; //out_dim-1;
         for (uint dim_i=0; dim_i<num_dims-1; dim_i++) {
           if (dim_i == 0) {
             size_lbmems *= img_dims[dim_i] / in_dims[dim_i];
           } else {
-            size_lbmems *= (img_dims[dim_i] - (out_dims[dim_i]-in_dims[dim_i])) / in_dims[dim_i];
+            size_lbmems *= img_dims[dim_i] / in_dims[dim_i];
+            //size_lbmems *= (img_dims[dim_i] - (out_dims[dim_i]-in_dims[dim_i])) / in_dims[dim_i];
```
leonardt: Do we need this comment?

src/libs/commonlib.cpp
```diff
@@ -1034,63 +1232,48 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
           }
         }
+
+        //size_lbmems = size_lbmems == 1 ? 0 : size_lbmems;
```
leonardt: Do we need this comment?

src/libs/commonlib.cpp
```diff
@@ -1177,18 +1353,28 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
               for (int dim_i=num_indices-1; dim_i>=0; --dim_i) {
                 if (dim_i == 0) {
                   // for last dimension, don't go to the end of register chain
-                  input_name += "." + to_string(out_dims[dim_i]-1 - indices[dim_i]);
+                  uint index_i = inverted_index(out_dims[dim_i], in_dims[dim_i], indices[dim_i]);
+                  input_name += "." + to_string(index_i);
+                  //input_name += "." + to_string(out_dims[dim_i]-1 - indices[dim_i]);
```
leonardt: do we need this comment?

src/libs/commonlib.cpp
```diff
@@ -1177,18 +1353,28 @@ Namespace* CoreIRLoadLibrary_commonlib(Context* c) {
...
                 } else {
                   input_name += "." + to_string(in_dims[dim_i]-1 - indices[dim_i]);
-
+                  //input_name += "." + to_string(indices[dim_i]);
```
leonardt: do we need this comment?

tests/ext/linebuffer_valid.cpp
```diff
@@ -31,10 +31,10 @@ int main() {
   // Define lb32 Module
   Type* lb32Type = c->Record({
 			{"in",in_type},
+      {"reset",c->BitIn()},
```
leonardt: Indentation looks off here

leonardt:
I just noticed the followup comment with the CGRAFlow passing, that can count as a test for this changeset, but in the future I would still appreciate seeing self contained unit tests for these generators included (so future changes can be easily tested without having to rely on the whole flow).

src/ir/headers/memories.hpp
jeffsetter: Code was modified to the existing generators, because after usage, they were not correct; these changes allow them to compile correctly. rom2 uses a port that is the same bitwdith as the data (even though this is more bits than necessary).

src/ir/headers/memories.hpp
jeffsetter: removed

Issue #627, opened 2018-10-06
Missing connections in coreir -> verilog


leonardt:
Input coreir json:
```
{"top":"global.Configurable",
"namespaces":{
  "global":{
    "modules":{
      "Configurable":{
        "type":["Record",[
          ["config_addr",["Array",32,"BitIn"]],
          ["config_data",["Array",32,"BitIn"]],
          ["config_en","BitIn"],
          ["O",["Array",32,"Bit"]],
          ["CLK",["Named","coreir.clkIn"]]
        ]],
        "instances":{
          "const_1_32":{
            "genref":"coreir.const",
            "genargs":{"width":["Int",32]},
            "modargs":{"value":[["BitVector",32],"32'h00000001"]}
          },
          "inst0":{
            "modref":"global.Register__has_ce_True__has_reset_False__has_async_reset__False__type_Bits__n_32"
          },
          "inst1":{
            "genref":"coreir.eq",
            "genargs":{"width":["Int",32]}
          },
          "inst2":{
            "modref":"corebit.and"
          }
        },
        "connections":[
          ["inst1.in1","const_1_32.out"],
          ["inst2.out","inst0.CE"],
          ["self.CLK","inst0.CLK"],
          ["self.config_data","inst0.I"],
          ["self.O","inst0.O"],
          ["self.config_addr","inst1.in0"],
          ["inst2.in0","inst1.out"],
          ["self.config_en","inst2.in1"]
        ]
      },
      "Mux2x32":{
        "type":["Record",[
          ["I0",["Array",32,"BitIn"]],
          ["I1",["Array",32,"BitIn"]],
          ["S","BitIn"],
          ["O",["Array",32,"Bit"]]
        ]],
        "instances":{
          "inst0":{
            "genref":"commonlib.muxn",
            "genargs":{"N":["Int",2], "width":["Int",32]}
          }
        },
        "connections":[
          ["self.I0","inst0.in.data.0"],
          ["self.I1","inst0.in.data.1"],
          ["self.S","inst0.in.sel.0"],
          ["self.O","inst0.out"]
        ]
      },
      "Register__has_ce_True__has_reset_False__has_async_reset__False__type_Bits__n_32":{
        "type":["Record",[
          ["I",["Array",32,"BitIn"]],
          ["O",["Array",32,"Bit"]],
          ["CLK",["Named","coreir.clkIn"]],
          ["CE","BitIn"]
        ]],
        "instances":{
          "inst0":{
[...]

PR #628, opened 2018-10-10
Fixed verilog global issue

rdaly525:


src/passes/analysis/vmodule.cpp
```diff
@@ -10,7 +10,7 @@ namespace VerilogNamespace {
 CoreIRVModule::CoreIRVModule(VModules* vmods, Module* m) : VModule(vmods) {
   Type2Ports(m->getType(),this->ports);
   assert(m->hasDef());
-  this->modname = m->getNamespace()->getName() + "_" + m->getLongName();
+  this->modname = m->getLongName();
```
leonardt: Does this mean we're not prepending the namespace names to the module names anymore?  Should this be checking that `getNamespace()` is not global and only prepending the namespace if it isn't? Or is that logic already being handled elsewhere?

rdaly525: The logic (should be) if you specifiy Metadata.verilog.prefix, it will prepend the specified prefix. If you do not specify, it will just use the module name.

rdaly525: Okay, that logic is broken for modules which have a CoreIR definition (but not for modules which have a verilog definition). I do not think that this case is being hit in any of your tests, so this PR will fix your issues.

leonardt:
LGTM, just a question about the implementation/intended behavior as I don't know the context of the change that well.

garnet and fault are still blocked by this issue https://github.com/rdaly525/coreir/issues/627

src/passes/analysis/vmodule.cpp
leonardt: Got it

PR #645, opened 2018-12-07
Add logging

rsetaluri:
Just a design proposal. Happy to iterate w some feedback. We can also explore the open source header-only libs.


rsetaluri:
@rdaly525 done.

PR #652, opened 2019-01-15
Simtarget

rdaly525:
depends on #649 which moves some APIs (common utils) from the simulator sub director to the coreir subdirector.

This PR separates coreir simulator code from coreir. One should now be able to install coreir separately from the simulator. This means there is a different include, shared lib, and binary for the simulator. This is a first step into cmakifying the coreir repo.

@leonardt, since the simulator now has a separate c shared lib, this will likely affect the APIs regarding simulator code from pycoreir. It is now called "coreirsim-c" 

Sorry for the insane amount of changes, but to summarize it:

src/ir/Makefile: Now does not use src/simulator/* to create libcoreir
src/simulator/*: creates its own libcoreirsim
src/pases/*: Now has no simulator dependencies
src/coreir-c: Separates the coreir-c lib from the coreirsim-c lib.
include/coreir.h: Does not include simulator headers. Automatically includes common util APIs
include/coreirsim.h: New generic include file for all the simulator includes.
include/coreir-c/* separates c api for coreir and simulator (coreir does not have any simulator dependency)
include/common/* common util APIs for coreir
Makefile: Now installs both coreir and coreirsim


src/binary/Makefile
```diff
@@ -3,7 +3,9 @@
 HOME = ../..
 INCS = -I$(HOME)/include -I.
 LPATH = -L$(HOME)/lib
-LIBS =  -Wl,-rpath,$(HOME)/lib -lcoreir -ldl -lcoreir-commonlib
+#TODO commonlib should not be a dependency here
```
leonardt: Could open a github issue to track this so it doesn't get lost

Issue #662, opened 2019-01-30
Coreir_slice invalid default parameterization


alexcarsello:
The following module is present in the generated verilog for the CGRA. As you can see the default parameterization is invalid, and Synopsys DC errors out because of this.
```verilog
module coreir_slice #(parameter hi=1, parameter lo=1, parameter width=1) (
  input [width-1:0] in,
  output [hi-lo-1:0] out
);
  assign out = in[hi-1:lo];

endmodule  // coreir_slice`
```
Errors:
`output [-1:0] out`
`assign out = in[0:1];`

rdaly525:
Tracked by #663

rdaly525:
@alexcarsello Can you confirm this is fixed in master?

alexcarsello:
@rdaly525 yes it's fixed now. Thanks.

PR #663, opened 2019-01-30
Slicefix

rdaly525:
running:
`coreir -i examples/concat.json -o out.v`
Shows the default values for the verilog slice module are viable


rdaly525:
Fixes #662

PR #666, opened 2019-02-02
Add verilog debug for stateful coreir primitives

leonardt:
Adds a `vjson` field `"verilog_debug"` which contains the definition of various coreir primtives including the `/*verilator public*/` annotations. I'm not sure if there's a more elegant way to do this, but it got the job done. The idea is that by default, we don't want these annotations because they force verilator to not inline signals, which can hurt performance.


leonardt:
Refactored this to just add a new field to the metadata `"debug_definition"`. This simplifies the logic, but still requires us to maintain two versions of the definition, which isn't great if we have to change them (need to remember to update it in two places). Is there a clean way to factor this code so we can just append the "/*verilator public*/" to the base definition?

Issue #672, opened 2019-02-06
Bugs in verilog output


jeffsetter:
The following generated verilog has two bugs in generation:
1) VTop does not have a clk signal (line 568 of uncorrected)
2) Invalid verilog syntax (line 48 of uncorrected)
   [uncorrected is "assign out = 1b'value;" ]

CoreIR json:
[conv_1_2_design_top.txt](https://github.com/rdaly525/coreir/files/2837610/conv_1_2_design_top.txt)
Uncorrected generated verilog:
[conv_1_2_top_uncorrected.txt](https://github.com/rdaly525/coreir/files/2837611/conv_1_2_top_uncorrected.txt)
Corrected verilog:
[conv_1_2_top.txt](https://github.com/rdaly525/coreir/files/2837613/conv_1_2_top.txt)

rdaly525:
Number 2 is fixed.

PR #673, opened 2019-02-06
removed the bitwidth prefix for the verilog constant

rdaly525:
This fixes bad verilog syntax when using bit constants.

This was changed a few months ago, but I do not recall why it was changed originally. I think it had something to do with the verilog inlining logic.

### DIFF
--- a/include/coreir/definitions/corebitVerilog.hpp
+++ b/include/coreir/definitions/corebitVerilog.hpp
@@ -16,7 +16,7 @@ void CoreIRLoadVerilog_corebit(Context* c) {
     {"other",{
       {"mux","sel ? in1 : in0"},
       {"concat","{in0, in1}"},
-      {"const","1b'value"},
+      {"const","value"},
       {"term",""},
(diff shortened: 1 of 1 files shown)


rdaly525:
Fixed #672

leonardt:
It looks like the inlining tests are failing, they aren't triggering an error code on travis though. I will investigate. IIRC, this was required because when a constant was inlined into a module definition, the parameterization of the bit width was no longer apparent because it was defined the module interface (the width of the output port) so you just got a constant with no width specifier. 

Here's an example where the `CIN` bit is zero extended, which the inlining logic collapses into a single verilog `concat`. 

```
assign O[7:0] = {1b'0, 1b'0, 1b'0, 1b'0, 1b'0, 1b'0, 1b'0, CIN} + I0 + I1;
```

Before, we were just getting `0`s. We can update the inlining logic to insert the required bit width specifier if this definition is causing issues for other tools.

Also, it looks like the inlining tests are not causing the travis test to fail, even though they are failing. I'll open an issue to track.

PR #687, opened 2019-02-14
Connfix

rdaly525:
This does the following:
Changes a few unordered_map/unordered_set to map/set
Rename Connection Comparison classes to be "Fast" or "Consistent".
Created an API to get sorted connections (Which should always be consistent)
Fixed a bug where verilog generation was always trying to inline. (It is possible that verilog inlining is still not deterministic)

### DIFF
--- a/include/coreir/ir/fwd_declare.h
+++ b/include/coreir/ir/fwd_declare.h
@@ -139,16 +139,17 @@ typedef std::pair<Wireable*,Wireable*> Connection;
 typedef std::vector<std::pair<Wireable*,Wireable*>> LocalConnections;
 
 
-//Comparison classes for caches
-class ConnectionComp {
+//Comparison classes
+
+//This will do fast comparisons on Connections, but makes no guarentees on being the same between different runs of the program
+class ConnectionCompFast {
   public:
-    static bool SPComp(const SelectPath& l, const SelectPath& r);
     bool operator() (const Connection& l, const Connection& r) const;
 };
 
-class ConnectionStrComp {
+//This will do slow comparisons on Connections but makes guarentee on being consistent between different runs
+class ConnectionCompConsistent {
   public:
-    static bool SPComp(const SelectPath& l, const SelectPath& r);
     bool operator() (const Connection& l, const Connection& r) const;
 };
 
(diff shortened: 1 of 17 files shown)


include/coreir/ir/fwd_declare.h
```diff
@@ -141,16 +139,17 @@ typedef std::pair<Wireable*,Wireable*> Connection;
...
     bool operator() (const Connection& l, const Connection& r) const;
 };
 
-class ConnectionStrComp {
+//This will do slow comparisons on Connections but makes guarentee on being consistent between different runs
+class ConnectionCompConsistent {
   public:
-    static bool SPComp(const SelectPath& l, const SelectPath& r);
     bool operator() (const Connection& l, const Connection& r) const;
 };
```
leonardt: Interesting idea. I think it makes sense to have two modes, one for testing/debugging, one for performance. We may want to do a design review and think about how we can fit his pattern into the overall system.

include/coreir/ir/typecache.h
```diff
@@ -3,6 +3,8 @@
 
 #include "fwd_declare.h"
 
+#include <unordered_map>
+
```
leonardt: Do you still need this include? (Not sure if there's other code in the file that uses it, but looks like you removed two below

tests/binary/gold/concat.v
```diff
@@ -47,16 +47,8 @@ module concats (
...
-  assign cc0__in0[3:0] = s0__out[3:0];
-
-  assign cc0__in1[11:0] = s1__out[11:0];
-
   assign out[15:0] = cc0__out[15:0];
 
-  assign s0__in[15:0] = in[15:0];
-
-  assign s1__in[15:0] = in[15:0];
-
```
leonardt: Wait, should these connections be missing?

rdaly525: Yeah, there is definitely still a bug here.

include/coreir/ir/fwd_declare.h
rdaly525: This was not quite the intended use case but would be interesting. I intend to have all Connection comparisons be "fast" in the core part of the IR but then when serializing (to verilog or CoreIR or to anything), using the slow deterministic one. The deterministic one is also available if a pass needs to use connections in a map or set deterministically.

PR #702, opened 2019-02-28
added the functions inlineInstance, getModParams, and newNamespace to…

rdaly525:
… the c api


rdaly525:
Totally agree about the bindings.

PR #704, opened 2019-02-28
Minor Fixes

jeffsetter:
Originally contained fake floating point ops in commonlib, but now removed those unneeded ops.

Some useful fixes are present to be merged in:
- removed erroneous  counter creation in linebuffer
- fixed verilog tb 
- added ram with 16 bit port to memories


jeffsetter:
Note that this merge is needed to bump master, which is passing CGRAFlow:
https://travis-ci.org/StanfordAHA/CGRAFlow/jobs/500064515

PR #705, opened 2019-03-02
Ice40 test

splhack:
- Split out and generalize verilog assert code from blackbox_verilog.cpp to assert_pass.h.
  - it allows to verify other pass, e.g. JSON `assertPassEq<Passes::CoreIRJson>(c, "golden.json");`
- Verify ice40 RAM and PLL Verilog code which are generated from JSON


splhack:
sure, moved😉

PR #654, opened 2019-01-17
Cmake

rdaly525:
This PR is Cmakeifying CoreIR. 

Tests are *not* being done using cmake currently. 

A large chunk of the files changed are due to fixing warnings when building the release version of cmake. These are fixed by changing assert -> ASSERT and by using a new coreir_unreachable() macro.

PR #706, opened 2019-03-02
[Proposal] Use Backward for printing pretty stack trace

splhack:
The current ASSERT stack trace is not demangled, so it is hard to debug. This proposal is introducing https://github.com/bombela/backward-cpp to print pretty stack trace instead. This PR uses libdw from elfutils because it supports Linux and macOS (I think.)

Sample output on unit test.

    Stack trace (most recent call last):
    #6    Object "", at 0xffffffffffffffff, in
    #5    Object "/home/kaz/coreir/tests/unit/build/blackbox_verilog", at 0x56310e758419, in _start
    #4    Source "../csu/libc-start.c", line 310, in __libc_start_main
    #3    Source "/home/kaz/coreir/tests/unit/blackbox_verilog.cpp", line 41, in main
             38: }
             39:
             40: int main() {
          >  41:   testBlackboxVerilog();
             42: }
    #2    Source "/home/kaz/coreir/tests/unit/blackbox_verilog.cpp", line 34, in testBlackboxVerilog
             31:   std::ifstream golden_stream("blackbox_verilog_golden.v");
             32:   std::string golden((std::istreambuf_iterator<char>(golden_stream)),
             33:                      std::istreambuf_iterator<char>());
          >  34:   ASSERT(golden == verilog,
             35:          "Expected '" + golden + "' but got '" + verilog + "'");
             36:
             37:   deleteContext(c);
    #1    Source "../../include/coreir/tools/backward.hpp", line 751, in load_here

### DIFF
--- a/include/coreir/ir/fwd_declare.h
+++ b/include/coreir/ir/fwd_declare.h
@@ -16,7 +16,24 @@
 
 #include <execinfo.h>
 
-#define ASSERT(C,MSG) \
+#if BACKWARD_HAS_DW == 1
+
+# include "coreir/tools/backward.hpp"
+# define ASSERT(C,MSG) \
+  do { \
+    if (!(C)) { \
+      std::cerr << "ERROR: " << MSG << std::endl << std::endl; \
+      backward::StackTrace st; \
+      backward::Printer p; \
+      st.load_here(20); \
+      p.print(st, stderr); \
+      exit(1); \
+    } \
+  } while(0)
+
+#else
+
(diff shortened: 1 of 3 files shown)


rdaly525:
This looks great! I just want to confirm that if someone does not have libdw, it will still work correctly.

splhack:
Yeah it works without libdw (less information though, just the same as the current stack trace).
And CMakefying(#654) would help to enable linking libdw by user easily!

PR #725, opened 2019-04-23
Adds more operators to floating point

jeffsetter:
add flr, ceil, min, max, neq


jeffsetter:
Passes CGRAFlow:
https://travis-ci.org/StanfordAHA/CGRAFlow/builds/524212024

PR #729, opened 2019-04-25
Support static linking of simulator

leonardt:


leonardt:
This doesn't seem to resolve the issue, will reopen when I have a working solution.

PR #731, opened 2019-04-26
added funtion to get metadata for modules

rdaly525:


src/coreir-c/coreir-c.cpp
```diff
@@ -257,6 +257,17 @@ extern "C" {
     }
   }
 
+  const char* COREModuleGetMetaData(COREModule* core_mod) {
+    Module* mod = rcast<Module*>(core_mod);
+    Context* c = mod->getContext();
+    string mstr = mod->getMetaData().dump();
+    std::size_t len = mstr.size() + 1;
+    char* cstr = c->newStringBuffer(len);
+    memcpy(cstr,mstr.c_str(),len);
```
leonardt: I think this memcpy should be `len + 1` so it copies the null terminator

rdaly525: The len is already defined to be the string size + 1. Not sure we need an extra one...

PR #732, opened 2019-04-26
Add BitInOut to C API

leonardt:


leonardt:
Also pushed some fixes required for compiling a simple `inout` <-> `inout` connection from magma. There was an issue in removebulkconnections (cherry-picked the fix from https://github.com/rdaly525/coreir/pull/607) and generating the direction of port strings in the verilog backend.

PR #738, opened 2019-05-04
bfloat division, equal and not equal

dillonhuff:
@jeffsetter @rdaly525 this PR includes bfloat division, equal, and not equal.

### DIFF
--- a/src/simulator/interpret.cpp
+++ b/src/simulator/interpret.cpp
@@ -1262,6 +1262,15 @@ namespace CoreIR {
           }
 
         });
+    } else if (opName == "float.eq") {
+      updateBitVecBinop(vd, [](const BitVec& l, const BitVec& r) {
+          return BitVector(1, l == r);
+        });
+      
+    } else if (opName == "float.neq") {
+      updateBitVecBinop(vd, [](const BitVec& l, const BitVec& r) {
+          return BitVector(1, l != r);
+        });
     } else {
       cout << "Unsupported node: " << wd.getWire()->toString() << " has operation name: " << opName << endl;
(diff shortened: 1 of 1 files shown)


dillonhuff:
@jeffsetter can you send me an example of what the division behavior should be? I don't know what the Halide bfloat emulation does, so I don't know what the input-output behavior is supposed to be.

jeffsetter:
Here are some values that I generated using the cpu implementation. Note that there may be a bug in this implementation, so let me know if something seems off.

Note that pi should be stored in bfloat as "4049" if looking at the bits in hexadecimal. All input values are translated from float to bfloat, perform division, and receive the value on the right.

Ran fp_pointwise on cpu
input(0,0) = 22   /pi = 7                                    (stored as 0x40e0)
input(1,0) = 10   /pi = 3.17188                          (stored as 0x404b)
input(2,0) = 3.17188   /pi = 1.0078125            (stored as 0x3f81)
input(3,0) = 3.17187   /pi = 1                            (stored as 0x3f80)
input(4,0) = 3.15625   /pi = 1
input(5,0) = 3.14063   /pi = 1
input(6,0) = 3.14062   /pi = 0.992188             (stored as 0x3f7e)
input(7,0) = 0.027451   /pi = 0.00866699      (stored as 0x3c0e)
input(8,0) = 0.0313726   /pi = 0.00994873  (stored as 0x3c23)
input(9,0) = 0.0352941   /pi = 0.0111694      (stored as 0x3c37)

rdaly525:
@dillonhuff, PR looks good. Ill merge, but can you create a FP div test?

PR #742, opened 2019-05-08
Includes more comparators: gt, ge, lt, le

dillonhuff:
@jeffsetter @rdaly525

### DIFF
--- a/src/simulator/interpret.cpp
+++ b/src/simulator/interpret.cpp
@@ -8,7 +8,7 @@ using namespace std;
 namespace CoreIR {
 
   BitVector truncateToBfloat(const BitVector& longRes) {
-    cout << "32 bit result = " << longRes << endl;
+    //cout << "32 bit result = " << longRes << endl;
     assert(longRes.bitLength() == 32);
 
     BitVector bRes(16, 0);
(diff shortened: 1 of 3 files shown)


src/simulator/interpret.cpp
```diff
@@ -1109,23 +1111,52 @@ namespace CoreIR {
...
+            int rv = r.to_type<int>();
 
-          float lf = bitCastToFloat(lv);
-          float rf = bitCastToFloat(rv);
+            float lf = bitCastToFloat(lv);
+            float rf = bitCastToFloat(rv);
 
-          cout << "lf = " << lf << endl;
-          cout << "rf = " << rf << endl;
+            cout << "lf = " << lf << endl;
```
rdaly525: Looks like there are some debug statements here and elsewhere in this file. Could you remove them?

dillonhuff:
@rdaly525 the printouts are now removed

dillonhuff:
@rdaly525 also why is float.abs a binary operation? I assumed it was short for absolute value.

rdaly525:
@dillonhuff, you are correct. That is a mistake and should be a unary

PR #743, opened 2019-05-10
Package recipes using modern CMake (required 3.8)

Bo-Yuan-Huang:
- Target-based recipe for in-source embedding
- Exporting the package (coreir library)


rdaly525:
@cdonovick, you are a bit more CMAKE savvy than I am. Could you do a review?

Kuree:
This will 100% break `pycoreir`...

rdaly525:
@Kuree, what issues do you foresee?

Kuree:
Per [PEP 513](https://www.python.org/dev/peps/pep-0513/#id40), the binary wheel has to satisfy the following conditions:
```
GLIBC <= 2.5
CXXABI <= 3.4.8
GLIBCXX <= 3.4.9
```
Due to the minimum requirement of `gcc-4.9`, we have to statically link the `glibc++`. This new PR removes the hacks that allow static linking in `pycoreir`.

Bo-Yuan-Huang:
Not an expert on Python binding, but it should be fine if it is about building a static library. 

https://github.com/Bo-Yuan-Huang/coreir/blob/e12b74d659a479e4763de50afe70132d2d7100a9/CMakeLists.txt#L23
One can enable static linking by turning the standard CMake build-in option ‘BUILD_SHARED_LIBS’ to OFF/FALSE.

@Kuree, is there other issue that will break the binding? I’d be glad to help!

Kuree:
Unfortunately that won't cut it. As I mentioned earlier, you need to statically link against `glibc` and `glibc++` to do so. You also need to be careful about the `D_GLIBCXX_USE_CXX11_ABI` flag. In my experience you need to turn that off to be compatible with modern C++ compilers.

One way to try out is to follow the build steps in `pycoreir` [here](https://github.com/leonardt/pycoreir/blob/master/.travis/setup.sh#L3). If you can produce a `manylinux` tagged wheels it means you're good to go. You need to run that in a docker container that's been modified from the official `manylinux` container. I think the easiest way is to fork `pycoreir` and change the remote where it clones `coreir` from.

Please let me know if you have any problems setting up the `pycoreir` build environment.

PR #757, opened 2019-06-07
Ubuffer

jeffsetter:
- Fix to mantle register to have reset priority
- Addition of Unified Buffer module interface


rdaly525:
@leonardt, will the change to the mantle register have any impact on Magma/mantle?

jeffsetter:
Mantle was modified a while back to have reset priority, but this was never reflected in CoreIR.
https://github.com/phanrahan/mantle/pull/150

PR #759, opened 2019-06-12
Dw

rdaly525:
This adds a library for instantiating dw FP mul and add.


rdaly525:
@leonardt, are there any necessary changes to pycoreir in order to get this to work?

PR #766, opened 2019-06-25
Add product list file output

rsetaluri:


rsetaluri:
Thanks, I'll add a test.

That's correct - using unique_ptr is actually a workaround for optional types not being available (they're not available til c++17 and we're using c++11). In fact, it's just a safer way of using the traditional hack of using a pointer as optional type (0 <--> None, anything else <-->Some). The alternative would be to upgrade to c++17, which I would actually suggest we do at some point :).

rsetaluri:
@leonardt included tests

PR #769, opened 2019-07-09
Simulator plugin

dillonhuff:
This is not done yet, but here is a running PR to track the simulator plugin.

### DIFF
--- a/src/passes/transform/fold_constants.cpp
+++ b/src/passes/transform/fold_constants.cpp
@@ -374,7 +374,102 @@ namespace CoreIR {
                        instPT->sel("in")->sel("out"));
           inlineInstance(instPT);
             
+        } else if (sigValue0.has_value()) {
+          BitVec sigVal0 = sigValue0.get_value();
+
+          if ((sigVal0.bitLength() == 1) && (sigVal0 == BitVec(1, 0))) {
+            cout << "Should remove and with 0" << endl;
+            //BitVec res = sigVal; //sigVal0 & sigVal1;
+
+            // uint inWidth =
+            //   inst->getModuleRef()->getGenArgs().at("width")->get<int>();
+
+            // assert(((uint) sigVal0.bitLength()) == inWidth);
+            // assert(((uint) sigVal1.bitLength()) == inWidth);
+            // assert(((uint) res.bitLength()) == inWidth);
+
+            // auto newConst =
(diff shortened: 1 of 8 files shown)


dillonhuff:
@rdaly525 @jeffsetter There is now a stub example of how to add a plugin in the unit tests of the interpreter. See the this PRs file changes for details.

src/passes/transform/fold_constants.cpp
```diff
@@ -374,8 +374,103 @@ namespace CoreIR {
                        instPT->sel("in")->sel("out"));
           inlineInstance(instPT);
             
+        } else if (sigValue0.has_value()) {
+          BitVec sigVal0 = sigValue0.get_value();
+
+          if ((sigVal0.bitLength() == 1) && (sigVal0 == BitVec(1, 0))) {
+            cout << "Should remove and with 0" << endl;
```
rdaly525: What is this comment? If it should error, then you should use ASSERT, otherwise there should be nothing to cout

src/passes/transform/fold_constants.cpp
```diff
@@ -374,8 +374,103 @@ namespace CoreIR {
...
+
+            Instance* instPT = addPassthrough(inst, "_inline_or_PT");
+            Select* replacement = newConst->sel("out");
+
+            def->removeInstance(inst);
+            def->connect(replacement,
+                         instPT->sel("in")->sel("out"));
+            inlineInstance(instPT);
+            
+            // auto recInstances = getReceiverSelects(inst);
```
rdaly525: Please remove commented code

src/passes/transform/fold_constants.cpp
```diff
@@ -474,6 +569,45 @@ namespace CoreIR {
...
+        //   Select* replacement = newConst->sel("out");
+
+        //   def->removeInstance(inst);
+        //   def->connect(replacement,
+        //                instPT->sel("in")->sel("out"));
+        //   inlineInstance(instPT);
+
+        // }
+
+        cout << "Not folding slice" << endl;
```
rdaly525: Similarly, if this is an error, use ASSERT. if you want debug statements, you can use:

LOG(INFO) << "comments";

Check out src/binary/coreir.cpp for an example

dillonhuff:
@rdaly525 just removed a bunch of comments and printouts.

PR #775, opened 2019-07-27
Another rom fix

dillonhuff:
@jeffsetter this should fix the rom issue. I have added a unit test of ROM2 that is now passing.


rdaly525:
@jeffsetter, can you confirm?

jeffsetter:
I would suggest removing some of the printouts (setting output of slice to...).

Other than that, it's working as expected!

dillonhuff:
@jeffsetter just removed the printouts.

Trying to change the formatting of the RAM is creating a bunch of errors. If this PR is working can we just merge it and do the formatting change in another request?

dillonhuff:
@jeffsetter just fixed the errors and changed the ram init format

PR #777, opened 2019-07-31
Adds support for inout connections

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/776 so pulled into that branch. This should either be merged first, or repointed to master before merging.

This adds initial support for inouts in the new verilog backend

### DIFF
--- a/tests/gtest/intermediate_connection_golden.v
+++ b/tests/gtest/intermediate_connection_golden.v
@@ -1,12 +1,18 @@
 // This is an important comment for foo!!
-module foo (input I, output O);
+module foo (input I, inout IO, output O);
     assign O = I;
+    assign IO = I;
 endmodule
-module top (input I, output O);
+module top (input I, inout IO0, inout IO1, output O);
+wire inst0_IO;
 wire inst0_O;
+wire inst1_IO;
 wire inst1_O;
-foo inst0(.I(I), .O(inst0_O));
-foo inst1(.I(inst0_O), .O(inst1_O));
+foo inst0(.I(I), .IO(inst0_IO), .O(inst0_O));
+foo inst1(.I(inst0_O), .IO(inst1_IO), .O(inst1_O));
 assign O = inst1_O;
+assign inst0_IO = inst1_IO;
+assign IO0 = inst0_IO;
+assign inst0_IO = IO1;
 endmodule
 
--- a/include/coreir/passes/analysis/verilog.h
+++ b/include/coreir/passes/analysis/verilog.h
@@ -1,10 +1,10 @@
 #ifndef COREIR_VERILOG_HPP_
 #define COREIR_VERILOG_HPP_
 
-#include <memory>
-#include <ostream>
 #include "coreir.h"
 #include "verilogAST.hpp"
+#include <memory>
+#include <ostream>
 
 namespace vAST = verilogAST;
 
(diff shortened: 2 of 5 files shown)


tests/gtest/intermediate_connection_golden.v
```diff
@@ -1,12 +1,18 @@
...
+wire inst1_IO;
 wire inst1_O;
-foo inst0(.I(I), .O(inst0_O));
-foo inst1(.I(inst0_O), .O(inst1_O));
+foo inst0(.I(I), .IO(inst0_IO), .O(inst0_O));
+foo inst1(.I(inst0_O), .IO(inst1_IO), .O(inst1_O));
 assign O = inst1_O;
+assign inst0_IO = inst1_IO;
+assign IO0 = inst0_IO;
+assign inst0_IO = IO1;
```
leonardt: @rdaly525 do you know if this is legal verilog? Basically, I wanted to cover the various cases inouts could be connected. Here we're assigning two drivers to `inst0_I0` which is an inout (and both the drivers are inouts). It lints in verilator, but not sure if that catches all problems.

rdaly525: inouts are a real annoyance in verilog. I wrote an email to you a while ago (which I cannot find right now) that lays out all the weirdness with inout.

The cases with inout are as follows:
1) Routing inouts between instances/module interface
2) Reading from inouts
3) Writing to inouts

for (1) you cannot have any assignments occur since verilog does not support an operator for bidirectional assign. The only way to route them is to name the intermediate wire something, and connect that intermediate wire to both the instance/interface ports

```
wire my_inout;
mymodule myinst0(.my_inout(my_inout));
mymodule myinst1(.my_inout(my_inout));
```

for (2) inouts can appear on the RHS of an assign easily. They are interpreted as just another wire whose value can be sampled from.

`assign my_normal_wire = my_inout;`

for (3), syntactically it is legal to have an inout on the lhs of multiple assignments. But this is an unused/unsafe usecase unless each assign statement is qualified as a ternary operation with high 'z' as the other option.

```
assign my_inout = my_pred ? val1 : 1'bz;
assign my_inout = ~my_pred ? val0 : 1'bz;
```

The following is syntactically legal, but in my opinion should throw a typechecking error
```
assign my_inout = val1;
assign my_inout = val2;
```
Since this essentially creates a 'short' whenever val1 != val2 and really does not have any valid use cases. Simulators will execute my_inout as an 'x' in this case.

leonardt: Okay, w.r.t. to the test, this is translating the description in CoreIR, so it seems like CoreIR could do a type checking pass to validate these properties (then we don't have to worry about generating that kind of illegal code since it won't be represented in the netlist). We'll have to update the test input once we implement that type check.

Basically, this is a semantics (static) issue, rather than a syntactic issue, so I think the logic to generate the verilog is okay and this issue will be handled either by the target or eventually when CoreIR implements the type checking pass.

leonardt: Actually, I'll to verify that case (1) is handled correctly.

rdaly525: There is no way to even construct case 2 or 3 in coreir since the CoreIR type system requires inouts to be connected to inouts. "tribuf" and "inbuf" need to be used to get the same behavior as 2 and 3.

leonardt:
Merging this as there are changed needs to unblock some downstream tools (fault), will take on the task of fleshing out the inout support, but it seems to work for the basic cases (the only user I know of is in garnet and this supports the use case for garnet).

PR #780, opened 2019-07-31
Add support for verilator_debug

leonardt:
Depends on #776, #777, #778, and #779.

This adds the logic to handle `verilator_debug` option which inserts the requisite verilator comments.  It's hacky, but I think this whole flow is hacky, so I vote that we add this temporary support so downstream tools are working, then refactor this entire flow (debug metadata).

Unfortunately I forgot to run clang-format on previous commits, so this includes some style fixes from clang-format. Here are the actual important changes:
* https://github.com/rdaly525/coreir/compare/patch-concat-order...patch-debug-definition?expand=1#diff-6c4fb4c3cc23c2c09bda419439646981R198-R202 (adds comment to ports)
* https://github.com/rdaly525/coreir/compare/patch-concat-order...patch-debug-definition?expand=1#diff-6c4fb4c3cc23c2c09bda419439646981R180-R186 uses debug version of definition if available

Also changes some helper functions to be methods so that we don't have to pass around shared fields (verilator_debug), eventually I'll move all the functions to be methods because I think that can simplify some of the code (don't have to pass around references to shared data structures).

### DIFF
--- a/include/coreir/passes/analysis/verilog.h
+++ b/include/coreir/passes/analysis/verilog.h
@@ -1,10 +1,10 @@
 #ifndef COREIR_VERILOG_HPP_
 #define COREIR_VERILOG_HPP_
 
-#include <memory>
-#include <ostream>
 #include "coreir.h"
 #include "verilogAST.hpp"
+#include <memory>
+#include <ostream>
 
 namespace vAST = verilogAST;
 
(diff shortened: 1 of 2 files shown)


leonardt:
@rdaly525 if the style changes are too annoying to parse, let me know and I can extract out the semantic changes and submit the style changes in a separate PR.

PR #785, opened 2019-08-09
Add osx build/release to travis flow

leonardt:
This changes the release flow to produce two tarballs named `coreir-{$OS}.tar.gz` where `$OS` is either `linux` or `mac`. It also runs the test suite under a macos environment in the travis builds.


rdaly525:
@leonardt, do you mind updating the INSTALL.md for instructions for installation instructions?

PR #801, opened 2019-10-07
Update flattentypes to work on declarations

leonardt:


leonardt:
Actually, this creates another error which I think is related:
```
  Def:
    Instances:
      base_module_inst0 : base_module
      ctr_module_inst0 : ctr_module
    Connections:
      base_module_inst0.baseIO.in0 <=> self.hier_baseIO_in0
      base_module_inst0.baseIO.out0 <=> self.hier_baseIO_out0
      ctr_module_inst0.ctr <=> self.hier_ctr

ERROR: {hier_module}.base_module_inst0.baseIO_in0 Is not connected
```

Perhaps the connection is not getting updated.

leonardt:
I added a warning in the flattentypes pass where it use to result in an error. I didn't add one to the instancegraph routine since that would lead to a double warning, perhaps this is okay since that's an internal method?

leonardt:
The original magma test now displays the warning as follows:
```
❯ pytest tests/test_type/test_tuple.py -k test_nested -s
/Users/lenny/miniconda3/lib/python3.7/site-packages/pysmt/walkers/generic.py:43: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  if len(nodetypes) == 1 and isinstance(nodetypes[0], collections.Iterable):
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.7.2, pytest-4.2.0, py-1.8.0, pluggy-0.9.0
rootdir: /Users/lenny/repos/magma, inifile: setup.cfg
plugins: parallel-0.0.9, cov-2.6.1, codestyle-1.4.0
collected 6 items / 5 deselected / 1 selected

tests/test_type/test_tuple.py /Users/lenny/repos/coreir/src/binary/coreir.cpp:188 Running Runningvpasses
/Users/lenny/repos/coreir/src/passes/transform/rungenerators.cpp:10 In Run Generators
/Users/lenny/repos/coreir/src/passes/transform/rungenerators.cpp:26 Done running generators
/Users/lenny/repos/coreir/src/passes/transform/flattentypes.cpp:102 Flatten type of generator or module with no definition, assumes definition follows the flatten types scheme, see https://github.com/rdaly525/coreir/issues/800 for more info
{global.base_module}.baseIO Is not a flattened type!
  Type is: {'in0':BitIn, 'out0':Bit}
```

PR #816, opened 2019-11-14
Fix handling of wrap generator in verilog backend

leonardt:
This fixes an issue where the Verilog backend tries to generator a
verilog parameter for the wrap generator definition.  The wrap generator
takes a coreir type to wrap and does not use this parameter in the
verilog code (unlike the other verilog generators).  This introduces a
special case in the backend to check for the name and skip the parameter
logic for the generator, but in the future we may want to chagne this to
use a metadata flag to generalize it.

Need to fix https://github.com/phanrahan/magma/issues/492


rdaly525:
Looks good to me.

PR #817, opened 2019-11-14
Fix handling of array of named types in verilog backend

leonardt:
The current check is too conservative and won't generate an array of named types when the named type corresponds to a Bit, this relaxes the check to reuse the logic to unwrap the named type to check its raw type.


leonardt:
bump

PR #819, opened 2019-12-05
Change CMake RPATH

hofstee:
If you use the existing settings for RPATH, coreir fails if you

```
make build
make install
make clean
```

due to `coreir: error while loading shared libraries: libcoreir.so: cannot open shared object file: No such file or directory`

This PR modifies RPATH according to the suggestions at https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling#always-full-rpath which relinks all libraries when installing so the RPATH resolves correctly when things are cleaned.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,8 +6,26 @@ file(GLOB SRC_FILES
 )
 
 #Create libcoreir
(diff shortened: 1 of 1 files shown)


hofstee:
Note this should also fix the LD_LIBRARY_PATH shenanigans people were running into before.

PR #820, opened 2019-12-05
Patch convert verilog connections

leonardt:
Fix bug discovered by https://github.com/leonardt/verilogAST-cpp/pull/20

The verilog backend was using an identifier node to store array selects (e.g. `x[0]`) instead of creating an explicit slice node (this triggered the escape identifier logic).  This fixes the logic to properly construct the slice node (and updates the code to handle the variant types)

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -356,7 +356,7 @@ convert_to_verilog_connection(Wireable *value) {
     select_path.pop_front();
   }
   std::string connection_name = "";
-  for (int i = 0; i < select_path.size(); i++) {
+  for (uint i = 0; i < select_path.size(); i++) {
     auto item = select_path[i];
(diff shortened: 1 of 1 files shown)


src/passes/analysis/verilog.cpp
```diff
@@ -3,6 +3,28 @@
 #include "coreir/tools/cxxopts.h"
 #include <fstream>
 
+// Unpack variant type and convert to parent type Expression
+std::unique_ptr<vAST::Expression>
+convert_to_expression(std::variant<std::unique_ptr<vAST::Identifier>,
+                                   std::unique_ptr<vAST::Index>> value) {
+  return std::visit([](auto &&value) -> std::unique_ptr<vAST::Expression> {
```
rsetaluri: great use of all the modern c++ :)

PR #823, opened 2019-12-20
Add inline verilog string metadata

leonardt:
This is a temporary solution for handling inline assertions inside magma modules.

We should investigate a more principled approach to passing through inline module assertions, but for now we just compile the assertions in python to verilog strings that are inlined into the module definition at the end.

An improvement would be to give CoreIR some representation of the string and allow it to interpolate names (e.g. tuple fields are replaced with the corresponding verilog name).  To do this, we would need to extend the system with some IR for representing verilog assertions (e.g. we could extend verilogAST to support these expressions, then introduce special nodes for CoreIR names that are eventually replaced with verilog names.

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -599,6 +599,10 @@ void Passes::Verilog::compileModule(Module *module) {
       body = compile_module_body(module->getType(),
                                  definition->getSortedConnections(),
                                  definition->getInstances());
+  if (module->getMetaData().count("inline_verilog") > 0) {
+    std::string inline_str = module->getMetaData()["inline_verilog"].get<std::string>();
+    body.push_back(std::make_unique<vAST::InlineVerilog>(inline_str));
+  }
 
   vAST::Parameters parameters = compile_params(module);
 
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -599,6 +599,8 @@ void Passes::Verilog::compileModule(Module *module) {
       body = compile_module_body(module->getType(),
                                  definition->getSortedConnections(),
                                  definition->getInstances());
+  // Temporary support for inline verilog
+  // See https://github.com/rdaly525/coreir/pull/823 for context
   if (module->getMetaData().count("inline_verilog") > 0) {
     std::string inline_str = module->getMetaData()["inline_verilog"].get<std::string>();
     body.push_back(std::make_unique<vAST::InlineVerilog>(inline_str));
--- a/tests/gtest/inline_verilog.json
+++ b/tests/gtest/inline_verilog.json
@@ -1,0 +1,34 @@
+{"top":"global.Main",
+"namespaces":{
+  "global":{
+    "modules":{
+      "FF":{
+        "type":["Record",[
+          ["I","BitIn"],
+          ["O","Bit"],
+          ["CLK",["Named","coreir.clkIn"]]
+        ]],
+        "metadata":{"verilog":{"verilog_string":"module FF(input I, output O, input CLK);\nalways @(posedge CLK) begin\n  O <= I;\nend\nendmodule"}}
+      },
+      "Main":{
+        "type":["Record",[
+          ["I","BitIn"],
+          ["O","Bit"],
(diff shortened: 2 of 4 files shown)


rdaly525:
How about having coreIR define a set of debugging nodes like 'assert' and 'print' which magma can compile into?

leonardt:
@rdaly525 that would work, but I think there's more to it than just that (ideally the nodes contain more than just verilog strings, e.g. there's some in memory representation of the assert/print contents, such as references to ports).  

Actually, now that I think about it, maybe we can use some sort of printf/str.format style syntax.

So something like `assert("assert property {{ @(posedge CLK) {%s} |-> ##3 {%s} }}", cls.I.valid, cls.O.ready)`.  Basically, magma can pass select paths and a string with format specifiers.  Coreir can convert the select path to verilog names then interpolate the strings using printf.

rdaly525:
That sounds good to me. You could have a json list of *args for the node’s parameter. Or alternatively have all the relevant printing signals be an inputs to the node. For assert the predicate should be a connected input.

leonardt:
There's also an issue with nested references (this is useful in verification, e.g. referring to some hierarchical sub-instance member path to a wire `foo.bar.d`).  I think we would need to extend coreir select paths to support selecting from instances

rdaly525:
Ah I see, good point. It might be interesting to allow connections in CoreIR to be able to select through the instance hierarchy. I would need to think of the implications. Otherwise if we are just passing the select path as a string/json, we could just come up with a way to encode that. Maybe something like (<instance_name>,"self",<rest of select path>)

rsetaluri:
I like Ross's idea of having CoreIR specific "debug" nodes, but agree it is a more involved change.

However, I'm not sure I like the idea of just being able to append arbitrary verilog. Working off your example in the PR, would verilog like this work? That way we have a separation.

```
module FF(input I, output O, input CLK);

always @(posedge CLK) begin
  O <= I;
end
endmodule
module Main (input CLK, input I, output O);
wire FF_inst0_O;
FF FF_inst0(.CLK(CLK), .I(I), .O(FF_inst0_O));
assign O = FF_inst0_O;

AssertHelper assert_helper_inst0(.I(I), .O(O));

endmodule

module AssertHelper(input I, input O);

assert property { @(posedge CLK) I |-> ##1 O };

endmodule```

If we have to stick to "inline verilog", I would change the name from `inline_verilog` to `verilog_suffix` or `verilog_addendum` or `appended_verilog`. "inline" is misleading.

**Comment by leonardt** [comment]
That's more involved, because we would need to know what signals to pass through to the helper module.  This code is appended to the end of the module body, so it can refer to internal signals.  It's really just so we have a stopgap solution for passing through inline verilog assertions for verification, I don't think this is a long term feature we want to support, especially for design purposes.  This is the desired input code: https://github.com/phanrahan/magma/pull/518/files#diff-75fcf4b4cd9167dbfbc261febcf0a4faR20-R22

The inline verilog is a bit akin to "inline assembly" (asm). The only difference here is that since there's no notion of "ordering statements" in the structural representation, we simply choose to generate the inline code at the end.  This way it can refer to any signals defined by the structural code.  The structural code cannot refer to any signals defined by the inline code, but that's okay (there isn't a mechanism for doing this).  

Semantically, appended or suffix verilog would work, but there's also a request for a stop gap solution for attaching Verilog after a module definition, so I'm worried it would conflate the naming with those properties.  So this verilog wouldn't be included "inline" in the definition, instead it is included after the `endmodule` statement.  The use case is to include verilog bind statements for attaching formal monitors to modules (these need to be done outside of a module, but are included with the module).

**Comment by rsetaluri** [comment]
Ok, that makes sense to me. It seems like a reasonable stop-gap solution that's not expected to be be supported long term. Perhaps a comment saying that above line 602 would be good.

File: src/passes/analysis/verilog.cpp [hunk 1]
```diff
@@ -599,6 +599,10 @@ void Passes::Verilog::compileModule(Module *module) {
       body = compile_module_body(module->getType(),
                                  definition->getSortedConnections(),
                                  definition->getInstances());
+  if (module->getMetaData().count("inline_verilog") > 0) {
```
rsetaluri: Add comment here saying this is temporary.

PR #827, opened 2020-01-13
Hotfix inline corebit

leonardt:
The inline logic was not properly handling the `corebit` primitives which are not generated modules.  This fixes the logic and adds tests to cover the cases when the primitive is a module and not a generated module.

Fixes https://github.com/phanrahan/magma/issues/529

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -107,6 +107,14 @@ bool can_inline_binary_op(CoreIR::Module *module, bool _inline) {
     return false;
 }
 
+std::unique_ptr<vAST::Expression> get_primitive_expr(CoreIR::Instance *instance) {
+    CoreIR::Module *module = instance->getModuleRef();
+    if (module->isGenerated()) {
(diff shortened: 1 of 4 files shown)


leonardt:
@rdaly525 we need this information to guide code generation (the inlining logic uses this to know whether or not it is using the binary op inliner, unary op inliner, etc..).  If we remove those attributes, we'd have to store them in a table in the verilog backend, which I'm not sure cleans up the code any better?

leonardt:
Basically, we can't remove this info, but we could move it somewhere else

rdaly525:
Doesn't all the information needed exist in the Module Type?

leonardt:
I'm not sure, can you give an example of the API to call to get that information? A pointer to where the header file defines them?

leonardt:
Also, I would suggest we address this issue in a separate PR, as this is a hot fix for a known bug, while I think you're asking for some refactoring which shouldn't affect functionality, so perhaps it's best to separate these changes.

PR #835, opened 2020-01-31
Fix verilator_debug inline bug

leonardt:
Fixes https://github.com/phanrahan/magma/issues/543, depends on https://github.com/leonardt/verilogAST-cpp/pull/38

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -460,13 +460,11 @@ Passes::Verilog::compilePorts(RecordType *record_type) {
     } else {
       ASSERT(false, "Not implemented for type = " + toString(type));
     }
-    std::unique_ptr<vAST::AbstractPort> port = std::make_unique<vAST::Port>(
-            process_decl(std::move(name), type), verilog_direction, vAST::WIRE);
+    std::unique_ptr<vAST::Port> port = std::make_unique<vAST::Port>(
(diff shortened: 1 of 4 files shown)


src/passes/analysis/verilog.cpp
```diff
@@ -461,16 +461,12 @@ Passes::Verilog::compilePorts(RecordType *record_type) {
...
     std::unique_ptr<vAST::Port> port = std::make_unique<vAST::Port>(
-            process_decl(std::move(name), type), verilog_direction, vAST::WIRE);
+            process_decl(std::move(name), type), verilog_direction,
+            vAST::WIRE);
     if (this->verilator_debug) {
-      // FIXME: Hack to get comment into port decl, we need to add support
-      // attaching comments to expressions
-      std::string port_str = port->toString();
-      port_str += "/*verilator public*/";
-      ports.push_back(std::make_unique<vAST::StringPort>(port_str));
```
rsetaluri: we should get rid of the `StringPort` class/node if not done already.

leonardt: Unfortunately StringPort is still used by the primitive code that uses a verilog string as the body and a list of strings for the interface ports (i.e. `verilog_json["interface"]` inside `compileStringBodyModule`), so we'll need to update that code path (mainly move all the primitives to use verilogAST) for that to be removed.

PR #839, opened 2020-02-06
Preserve port order in module instantiation

leonardt:
Updates for https://github.com/leonardt/verilogAST-cpp/pull/39 so that
the order of ports for module instances are preserved.  The old logic
used a std::map which uses lexical order, causing the output to sort
the ports based on the names.  Migrating to a std::vector preserve's
CoreIR's internal ordering of the ports (which is based on insertion
order)


rsetaluri:
See https://github.com/leonardt/verilogAST-cpp/pull/39. I think this is a good change but we should lock it into the spec, so that the generated verilog doesn't keep changing.

PR #840, opened 2020-02-12
Fix double delete

leonardt:
In this case, `argv` is an array of pointers that are references to the underlying c_string of a  `std::string`.  By calling `delete[]`, this implicitly invokes `delete` on the contents of the array (according to https://stackoverflow.com/questions/17344727/does-delete-call-destructors), but these contents are pointers owned by another `string` object.  I suspect this is related to https://github.com/StanfordAHA/garnet/issues/428, but the issue still exists so I'm working through it to see if there's more issues.


src/ir/passmanager.cpp
```diff
@@ -123,7 +123,10 @@ bool PassManager::runPass(Pass* p,vector<string>& pArgs) {
   if (argc > 1) {
     p->initialize(argc,argv);
   }
-  delete[] argv;
```
rsetaluri: we should just do the following:

```python
std::vector<char*> argv(argc);
for (auto& pArg : pArgs) {
  argv[i] = pArg.c_str();
}
```
then you don't have to deal w/ any `delete` business. There really shouldn't be raw `new`'s and `free`'s in the code.

(you may have to do some business w/ `const_cast<>` in the code above but I think it should work)

rdaly525: Lets just use the answer in this SO https://stackoverflow.com/questions/26032039/convert-vectorstring-into-char-c

rsetaluri: believe that's exactly my suggestion above :D

PR #841, opened 2020-02-12
Remove static variable

leonardt:
This fixes https://github.com/StanfordAHA/garnet/issues/428

I'm not exactly sure what the issue is, but I suspect it's related to the fact that a static member is being used in static linking context.  Using gdb backtrace lead me to this as the source of the double free, and changing it to use normal non-static logic resolves the issue, although we should probably update all the passes to use the same form (I'm not sure why it only happens with this one, and/or whether it will resolve the issue for others).


rsetaluri:
That's weird...wonder if it https://stackoverflow.com/questions/6714046/c-linux-double-destruction-of-static-variable-linking-symbols-overlap get's at the heart of it, i.e. it's linked statically and dynamically simultaneously somewhere

leonardt:
@rsetaluri That's my theory, particularly because CoreIR is combining a statically link libcoreir with dynamically loaded libraries like commonlib

rdaly525:
Ill keep the branch around for now in case anyone is still using it

rsetaluri:
@leonardt @rdaly525 if this unblocks people so be it, but we should *definitely* fix the real source of this issue, not just get rid of `static`'s

Issue #842, opened 2020-02-12
Change all static string ID for passes to non-static


rdaly525:
Fixes double free issue for #841 
This issue is to make the rest of the passes consistent with this fix.

PR #844, opened 2020-02-12
Remove static ID logic

leonardt:
Fixes https://github.com/rdaly525/coreir/issues/842

Seems like the main way this is used to conveniently refer to the name of a pass through the static class member of the type.  Without this, we have to pass both a template type `T` and an ID string to some functions.  Perhaps there's another way to do this without using a static class member? Or a way to do this safely in our static linking context?


rdaly525:
One thing is this will almost surely have impacts on dependencies (like halide, etc...) We should run GarnetFlow before merging this.

leonardt:
Closing in favor of #913

PR #845, opened 2020-02-14
Preserve Intermediate Wires

leonardt:
This updates the verilog backend to handle the inlining of Wire primitive modules such that it preserves the intermediate wire name.  Combined with https://github.com/phanrahan/magma/pull/584, this will enable users to pass through named temporary values to be generated as verilog wires.  The logic is only enabled in `--inline` mode.  It checks if we're inlining a Wire module, if so, we emit a wire named after the module instance and assign the input to the wire.  We replace all references to the module outputs with the wire instance name.  Finally, we use https://github.com/leonardt/verilogAST-cpp/pull/40 to prevent the inlining logic from removing the temporary wires by blacklisting the wire instance names.

Depends on https://github.com/leonardt/verilogAST-cpp/pull/40


src/passes/analysis/verilog.cpp
```diff
@@ -809,8 +843,8 @@ compile_module_body(RecordType *module_type,
     body.push_back(std::move(statement));
   }
   // Wire the outputs of the module and inout connections
-  assign_module_outputs(module_type, body, connection_map);
-  assign_inouts(definition->getSortedConnections(), body);
+  assign_module_outputs(module_type, body, connection_map, _inline);
+  assign_inouts(definition->getSortedConnections(), body, _inline);
```
leonardt: Should make these methods on the class so we can access the _inline member rather than having to pass them around.

tests/gtest/golds/bits_instance.v
```diff
@@ -0,0 +1,9 @@
+// Module `Foo` defined externally
+module Main (input [4:0] I, output [4:0] O);
+wire [4:0] Foo_inst0_O;
+wire [4:0] x;
+Foo Foo_inst0(.I(I), .O(Foo_inst0_O));
+assign x = Foo_inst0_O;
```
leonardt: Ideally we'd "inline" this x into the module instantiation, but I would say that's a low priority feature.

src/passes/analysis/verilog.cpp
rsetaluri: leave a TODO here for this

PR #848, opened 2020-02-26
Hack in support for async reset reg

leonardt:
In testing for the magmathon, I noticed that the interpreter doesn't support `coreir.reg_arst`.

This hacks around the basic functionality for our purposes, but now that I understand the code I think I can update it to support the general register features, but may not get time to do that until after the hackathon (this could be a good project)

### DIFF
--- a/src/simulator/interpret.cpp
+++ b/src/simulator/interpret.cpp
@@ -938,7 +938,7 @@ namespace CoreIR {
       });
     } else if ((opName == "coreir.const") || (opName == "corebit.const")) {
     } else if (opName == "corebit.term") {
-    } else if ((opName == "coreir.reg") || (opName == "corebit.reg")) {
+    } else if ((opName == "coreir.reg") || (opName == "corebit.reg") || (opName == "coreir.reg_arst")) {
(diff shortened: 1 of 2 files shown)


src/simulator/interpret.cpp
```diff
@@ -1234,30 +1234,48 @@ namespace CoreIR {
...
         (clkVal->value() == 1)) {
 
-      if (inSels.size() == 2) {
 
         setRegister(inst->toString(), bv1); //s1->getBits());
         ASSERT(same_representation(getRegister(inst->toString()),bv1),inst->toString() + " != " + toString(bv1)); //s1->getBits());
 
-      } else {
-        assert(inSels.size() == 3);
+      // TODO: Unused enable logic, remove?
```
leonardt: Since `ASSERT(inSels.size() == 2, ...)` was above, this code path wasn't being used, so I commented it out for now (I think enables are handled as explicit muxes?).  I can also remove it.

leonardt: I went ahead and removed this.

PR #850, opened 2020-03-05
proposed cse

rsetaluri:
@rdaly525 just staging so I can leave reviews 👍

### DIFF
--- a/src/passes/transform/cse.cpp
+++ b/src/passes/transform/cse.cpp
@@ -1,0 +1,174 @@
+#include "coreir/passes/transform/cse.h"
+
+using namespace std;
+using namespace CoreIR;
+
+// Desired CoreIR iterators/functional methods
+// breadth-first tree iterator on wireables
+// depth-first tree iterator on wireables
+// leaf-iterator on wireables (Only single bits)
+// breadth-first iteration on types
+// depth-first iterator on types
+// leaf iterator on types 
+// 'All' operator with a predicate
+// 'Any' operator with a predicate
+// 'Map' operator
+// 'Filter' operator (both a modify and a return new)
+
+
+//Some possible (untested) implementations of some of the functions
+template <typename Cont, typename Pred>
+Cont filterOver(const Cont &cont, Pred predicate) {
+  Cont result;
+  std::copy_if(cont.begin(), cont.end(), std::back_inserter(result), predicate);
+  return result
+}
+
+template <typename InCont, typename OutCont, typename Fun>
+OutCont mapOver(const InCont &cont, Fun fun) {
+  OutCont result;
+  std::transform(cont.begin(), cont.end(), std::back_inserter(result), fun);
+  return result
+}
+
+template <typename Cont, typename Pred>
+bool anyOf(const Cont &cont, Pred predicate) {
+  return std::any_of(cont.begin(), cont.end(), predicate);
+}
+
+template <typename Cont, typename Pred>
+bool allOf(const Cont &cont, Pred predicate) {
+  return std::all_of(cont.begin(), cont.end(), predicate);
+}
+
+
+//This is a pass that will do generalized common subexpression elimination.
+//a 'cse' is any time there are two instances of the same module which 
+//are driven by identical values.
+namespace CoreIR {
+
+  namespace Passes {
+
+    string Passes::CSE::ID = "cse";
+    bool CSE::runOnModule(Module* m) {
+      if (!m->hasDef()) {
+        return false;
+      }
+
+      //create map from instance kind to list of instances
+      std::map<Module*, std::vector<Instance*>> modules2Instances;
+      for (auto ipair: m->getInstances()) {
+        auto imod = instance->getModule();
+        if (modules2Instances.count(imod)) {
+          modules2Instances[imod] = {};
+        }
+        modules2Instances[imod].add(instance)
+      }
+      
+      bool modified = false;
+      for (auto mpair : modules2Instances) {
+        auto mod = mpair.first;
+        auto ilist = mpair.second;
+
+        //instance list needs to have at least 2 instances
+        if (ilist.size() < 2) {
+          continue;
+        }
+        
+        //Module cannot contain an inout port
+        auto it = leaf_iter(mod->getType());
+        bool has_inout = std::any_of(it.begin(), it.end(), [](Type* t) {
+          return t->isInOut();
+        });
+        if (has_inout) {
+          continue;
+        }
+
+        //Filter out any instances that have a connected mixed direction
+        ilist = filterOver(ilist, [](Instance* inst) {
+          auto iter = bfirst_iterator(inst);
+          bool has_mixed = std::any_of(iter.begin(), iter.end(), [](Wireable* w) {
+            return w->getConnectedWireables().size() > 0 &&
+              w->getType()->isMixed();
+          });
+        });
+
+        //Search for potential matches among the instance list
+        //pop first from list
+        //if has matches, add to matchlist
+        //  pop all matches from list as well
+        //This can be a helper function
+        auto getInputsAndDirivers = [](Instance* inst) {
+          auto iter = bfirst_iterator(inst);
+          auto inputs = filterOver(iter.begin(), iter.end(), [](Wireable* w) {
+            return w->getType()->isInput()
+                && w->getConnectedPorts().size() > 0;
+          });
+          auto drivers = mapOver(cur_inputs, [](Wireable* w) {
+            return w->getConnectedWireables()[0];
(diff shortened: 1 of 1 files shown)


src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
+#include "coreir/passes/transform/cse.h"
+
+using namespace std;
```
rsetaluri: don't do this!

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+using namespace CoreIR;
```
rsetaluri: don't need this! and also don't do this!

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+// Desired CoreIR iterators/functional methods
```
rsetaluri: better formatting here:

```c++
// Desired CoreIR iterators/functional methods:
//   * breadth-first tree iterator on wireables
//   * depth-first tree iterator on wireables
//   * leaf-iterator on wireables (Only single bits)
//   * breadth-first iteration on types
//   * depth-first iterator on types
//   * leaf iterator on types 
//   * 'all' operator with a predicate
//   * 'any' operator with a predicate
//   * 'map' operator
//   * 'filter' operator (both a modify and a return new)
```

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+// breadth-first iteration on types
+// depth-first iterator on types
+// leaf iterator on types 
+// 'All' operator with a predicate
+// 'Any' operator with a predicate
+// 'Map' operator
+// 'Filter' operator (both a modify and a return new)
+
+
+//Some possible (untested) implementations of some of the functions
```
rsetaluri: Space between `//` and first character of comment. End full line comments in `.`. Make sure to wrap at 80 chars.  Same for all comments below.

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+template <typename Cont, typename Pred>
```
rsetaluri: are `Cont`,  `Pred`, `Fun`, etc. standard abbreviations? It's totally fine if so and I'm not a domain expert (I can guess `Pred` is predicate and `Fun` is function but just looking at the function signature doesn't do much for me in terms of variables). And you have `Cont cont` but `Pred predicate`...keep it consistent between abbreviations and full names.

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+bool anyOf(const Cont &cont, Pred predicate) {
+  return std::any_of(cont.begin(), cont.end(), predicate);
+}
+
+template <typename Cont, typename Pred>
+bool allOf(const Cont &cont, Pred predicate) {
+  return std::all_of(cont.begin(), cont.end(), predicate);
+}
+
```
rsetaluri: nit: only 1 newline

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+// leaf-iterator on wireables (Only single bits)
+// breadth-first iteration on types
+// depth-first iterator on types
+// leaf iterator on types 
+// 'All' operator with a predicate
+// 'Any' operator with a predicate
+// 'Map' operator
+// 'Filter' operator (both a modify and a return new)
+
+
```
rsetaluri: nit: only one newline

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+  return std::all_of(cont.begin(), cont.end(), predicate);
+}
+
+
+//This is a pass that will do generalized common subexpression elimination.
+//a 'cse' is any time there are two instances of the same module which 
+//are driven by identical values.
+namespace CoreIR {
+
+  namespace Passes {
```
rsetaluri: don't indent nested namespaces:

```c++
namespace CoreIR {
namespace Passes {

// bunch of code here...

}  // namespace Passes
}  // namespace CoreIR
```

Notice the newline spacing too

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    string Passes::CSE::ID = "cse";
+    bool CSE::runOnModule(Module* m) {
```
rsetaluri: nit: newline between static var assignment and function

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+                mod->addConnection(srcSP, dstSP);
+              }
+            }
+            //Delete other instance (will also delete all connections)
+            mod->removeInstance(other);
+          });
+        }
+      }
+      return modified;
+    } //runOnModule
```
rsetaluri: nit: don't need `// runOnModule`

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+        for (auto matches : matchList) {
+          assert(matches.size() > 1);
+          //This one will be kept
+          auto inst = matches.pop_back();
+          auto iname = inst->getInstanceName();
+          for (auto other : matches) {
+            //find connected outputs
+            auto it = bfirst_iter(other);
+            auto outputs = filterOver(it.begin(), it.end(), [](Wireable* w) {
+              return w->getType()->isOutput()
```
rsetaluri: `&&` goes on first line

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+//This is a pass that will do generalized common subexpression elimination.
+//a 'cse' is any time there are two instances of the same module which 
+//are driven by identical values.
+namespace CoreIR {
+
+  namespace Passes {
+
+    string Passes::CSE::ID = "cse";
+    bool CSE::runOnModule(Module* m) {
+      if (!m->hasDef()) {
```
rsetaluri: a lot of these short one-line if's look better as a single line like:

```C++
if (...) continue;
if (...) return x;
```

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+        return false;
+      }
+
+      //create map from instance kind to list of instances
+      std::map<Module*, std::vector<Instance*>> modules2Instances;
+      for (auto ipair: m->getInstances()) {
```
rsetaluri: a lot of these names are not descriptive: `ipair`, `imod`, `iname`. (Again maybe it's understandable to the experts, but e.g. I couldn't really figure out what was going on. Code/names as documentation is always a good idea.)

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+  namespace Passes {
+
+    string Passes::CSE::ID = "cse";
+    bool CSE::runOnModule(Module* m) {
+      if (!m->hasDef()) {
+        return false;
+      }
+
+      //create map from instance kind to list of instances
+      std::map<Module*, std::vector<Instance*>> modules2Instances;
```
rsetaluri: an aside: this seems like a general operation...should it be factored out? is it used elsewhere already?

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
...
+// leaf iterator on types 
+// 'All' operator with a predicate
+// 'Any' operator with a predicate
+// 'Map' operator
+// 'Filter' operator (both a modify and a return new)
+
+
+//Some possible (untested) implementations of some of the functions
+template <typename Cont, typename Pred>
+Cont filterOver(const Cont &cont, Pred predicate) {
```
rsetaluri: I don't have access to the header file, so I can't tell but are these supposed to be externally available free functions? Or meant to be just for internal use to this CSE pass? If they're external then the should probably be in some more generic file (ala `util.hpp` or `functional.hpp`);  otherwise they should be in anonymous namespace:

```c++
namespace {

template<...>
Cont filterOver(...) {
  ...
}

...

}  // namespace
```

src/passes/transform/cse.cpp
```diff
@@ -0,0 +1,174 @@
+#include "coreir/passes/transform/cse.h"
```
rsetaluri: It took me a while to gather that cse => common subexpression elimination. I also see that it is part of the LLVM lexicon however.  A comment either here or in the header, that this is a common subexpression elimination pass would be useful. And exactly what types of subexpressions it can do would be useful for those who want to use the pass.

rdaly525:
Taking this down, as this was not the purpose of the review.

PR #852, opened 2020-03-17
Hierarchical select

leonardt:
This adds support for hierarchical select paths.  These changes do two things: add support to hierarchical selects to the Wireable code path, and add support for code generating hierarchical selects to the verilog backend.

The first use case is for passing debug/verification modules through the tool chain.  The initial flow will generate a separate module for each debug/verification node.  Internally the module will reference magma/coreir values as interface ports that are inlined into the verilog.  The module is instanced in the desired location, and the desired signals wired up to the instance.  This way, if coreir makes any changes to the circuit structure (e.g. optimizations) the connections to the debug/verification instance should be maintained and handle any changes in signal names.  For this to work, we need to be able to hierarchically connect signals to these debug nodes (e.g. wire up the internal signal of an instance to the debug node).  This feature is requested by our users.  

An extension to this logic might be to transform the CoreIR circuit to remove hierarchical selects and use the standard module/connection system (by inserting ports to pass these signals around), however I will leave this as future work since for the debug/verification case in verilog, we can simply code generate using Verilog's hierarchical attribute syntax.

Another extension will be to have CoreIR handle the renaming of references in the inline verilog.  For example, a module will have a 2-d array of tuples as an input.  In the debug/verification verilog, the user references a member of a tuple.  Ideally, we could stick in a placeholder symbol and have CoreIR interpolate the flattened name (perhaps a simple search/replace on a unique string).  Alternatively, if we represent the debug/verification logic directly in coreir as nodes, we can simply have the flattening and verilog backend handle this.

Depends on https://github.com/leonardt/verilogAST-cpp/pull/42


include/coreir/ir/wireable.h
```diff
@@ -128,12 +129,15 @@ class Select : public Wireable {
   protected :
     Wireable* parent;
     std::string selStr;
+    bool _isInstance;
```
rdaly525: Is this trying to differentiate between a hierarchical select and a normal select? I think the hierarchical select would be better as a separate class (inherited from Select). I am a little worried that this might break code which recursively goes through all the selects of an instance (or interface).

src/ir/module.cpp
```diff
@@ -134,4 +134,12 @@ bool Module::hasVerilogDef() {
   return false;
 }
 
+bool Module::canSel(std::string sel_str) {
```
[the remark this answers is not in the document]
leonardt: Convenience helper so I don’t have to check if a module def exists before checking the canSel function.

include/coreir/ir/wireable.h
leonardt: What would break? This provides the same interface for existing selects, so it shouldn’t break any existing code, it just adds another interface function.

src/ir/wireable.cpp
```diff
@@ -27,17 +27,23 @@ Wireable::~Wireable() {
...
+  Select* select = new Select(this->getContainer(), this, sel_str, type, is_instance);
+  selects[sel_str] = select;
+  return select;
+}
+
 Select* Wireable::sel(const std::string& selStr) {
   if (selects.count(selStr)) {
     return selects[selStr];
   }
+  if (isa<Instance>(this) &&
```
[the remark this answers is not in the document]
leonardt: I think self is a little awkward, maybe instA._instances_ or some other special but explicit keyword. Using self is confusing w.r.t. it referencing the interface in one context and instances in another context.

Is having the same name for an instances and a port desirable? I guess it’s technically okay because interface ports are prefixed by self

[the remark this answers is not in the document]
leonardt: I’ll move the logic into instance::sel (I originally did this but I think I ran into some issue, but I’ll reply if I come across it again)

include/coreir/ir/wireable.h
leonardt: i can see how much code it would take to change to add a new select class, my main concern is there’s a lot of code that uses Select so adding a new class to the code path will require a lot of changes to dispatch on the two types of selects, so we should decide whether it’s worth treating them separately versus treating a select as a single entity with a unified interface.

leonardt: i can see how much code it would take to change to add a new select class, my main concern is there’s a lot of code that uses Select so adding a new class to the code path will require a lot of changes to dispatch on the two types of selects, so we should decide whether it’s worth treating them separately versus treating a select as a single entity with a unified interface.

src/ir/wireable.cpp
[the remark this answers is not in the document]
rdaly525: I think _instances_ is also slightly semantically wrong as well. We want a word that is saying "select the inside of this current instance". Also something concise would be nice for serializing long hierarchical select paths.

include/coreir/ir/wireable.h
rdaly525: The thing I am concerned about is the following:

Lets say I have Module A instantiating Module B which instantiates Module C
Additionally lets say within A, I have a connection referring to the instance of C within A's instance of B.

During an optimization pass of B I do not want to have access to the select of C and the fact that some unknown Module (A) is using that select.

src/ir/wireable.cpp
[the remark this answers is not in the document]
leonardt: Maybe a separator different than dot?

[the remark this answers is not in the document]
rdaly525: I would be fine with that. Semicolon perhaps.

"instA;in0" vs "instA.in0"
and for the select path:
("instA", ";", "in0") vs ("instA", "in0")

include/coreir/ir/wireable.h
leonardt: That could be achieved by filtering the selects (in the function to gather connected wires led) rather than encoding it in the class system. There might be cases when you do want to know about the connection (e.g. if you’re removing the port you’ll probably want to remove that connection in another context?) that brings up a question about how a hierarchical selects will affect optimization passes

leonardt: Another option is to have magma lift these hierarchical selects into the interface and drop support for it in coreir directly, which may make sense given the above issue with optimization passes (cc @rsetaluri, any thoughts? We originally discussed this as something that might be useful in coreir, but it may introduce semantic complexity we dint want to have in the IR)

rdaly525: I think this feature is a good thing to have in CoreIR and by lifting it into the interface directly, you are losing information that could be passed down and are modifying interfaces unnecessarily.

I am just trying to enumerate possible issues with this.

rdaly525: Okay I think the filtering would work, but there needs to be an update to the logic so that any Select from a hierarchical select is also a hierarchical select. Also making sure that the connection does not appear in the Module with the hierarchical select.

leonardt: Another thought, do we want to enable connecting to inputs hierarchically? For example, one use case is to monitor an internal signal, so we can reference an output and send it into a debug/verification node.  Do we want to support users wiring to inputs hierarchically? This may conflict with our current definition semantics, where everything must be driven or termed.  essentially, you would leave an input undriven, and it could be wired up from the instance location.  In this case, we'd want to run the undriven check at the end on the entire hierarchy rather than per module (not sure if this is what is already done).  Otherwise, we should update the connect logic to make sure hierarchical selects only reference outputs.  I believe the current use case only wants to monitor signals, so this may be a good simplification to start with.

rdaly525: Yeah, I think we should constrain hierarchical selects to outputs for now (as that is the pressing use case). Driving hierarchical selects opens up a larger can of worms and we should do a review of tools that allow this.

leonardt: I confirmed that the users only want to monitor signals, and that they aren't aware of any patterns for driving signals this way, so I don't think we need to support it (at least until a valid use case comes up).

Perhaps for now we can treat these a special selects for debug nodes? That might help avoid having them interfere with standard "RTL" code/nodes

rdaly525: That could work. How about for now to simplify this just instantiate debug nodes, do not connect them to anything, and store the hierarchical select path as the debug node's metadata or modargs.

leonardt: yea that seems like a good idea and less invasive for current code paths.  I'll resubmit a new PR with this change as it will require a different overall approach.

leonardt: Actually, there's still the issue of maintaining the select path when optimizations are run (e.g. what if a monitored signal gets renamed in an optimization, ideally we'd still be able to connect to that signal).  That's why I wanted to piggy back on the connection logic, since ideally that'd be kept up to date

leonardt: So in some sense, optimizations should be aware of these "external" hierarchical selects, if just to make sure that their connections are preserved

rdaly525: Yeah, I understand the concern.

I think the biggest constraint is that CoreIR assumes that Modules and module Defs are completely independent of the outside world (namely other modules that could instantiate them). Therefore connections should always be between two wireables within the same parent Module.

Here is a half baked idea that I think could still work with your constraints. What if we have some sort of "portal" node that can go between module boundaries without going through the interface.

So for example in the A instances B which instances C.
You can instantiate one end of the portal in A (and connect it to the debug node), then also instantiate a portal node in B and (connect it to the appropriate signal within B). Any optimizations in B (even renaming signals) should not effect the result. The only issue is that optimizations in A (like renaming the instance of B) might pose a problem.

leonardt: I was thinking about this some more, I think for debug/verification purposes, having a reference to a signal should prevent the signal from being optimized out.  So a subtract could get turned into an an add with invert, so long as the original signal reference now points to the output of the add.  I guess with common sub expression elimination, if the signal mapped to the common sub expression output that would be okay too.  Is there ever a case where a signal might be removed entirely? (Without a new mapping)

rdaly525: Yeah, as long as there is something connected to the signal you care about, then any optimizations should still work. There is a chance that a particular signal might get optimized away. For example if you are trying to poll the output of an Add(a, 0). If this gets optimized away without the module knowing that something is connected to its output, then the reference is lost.

rdaly525: This is why if there is something that the desired signal is connected to a portal or term node of some sorts, then the reference will be updated appropriately during any optimizations.

src/ir/wireable.cpp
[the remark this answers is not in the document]
leonardt: updated it to use `;` for instances

include/coreir/ir/wireable.h
leonardt: I think for now, I will see if I can add a separate class for InstanceSelect otherwise stick to the boolean flag on the Select type, and use a symbol table to handle renaming by optimizations downstream, and if needed we can add a filter for instance selects when gathering connected wireables.

CMakeLists.txt
```diff
@@ -78,3 +78,7 @@ add_test(NAME test_verilog COMMAND test_verilog WORKING_DIRECTORY ${CMAKE_SOURCE
 add_executable(test_inline_wires tests/gtest/test_inline_wires.cpp)
 target_link_libraries(test_inline_wires gtest_main coreir coreir-commonlib)
 add_test(NAME test_inline_wires COMMAND test_inline_wires WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests/gtest)
+
```
rdaly525: I must have missed this last time, but why are specific tests specified in the base CMakeLists.txt? Can we move this to a CMakeLists.txt in the test directory?

leonardt:
Added test for name aliasing, for what it's worth, it produced invalid verilog (according to verilator) with this error:
```
 verilator --lint-only tests/gtest/golds/hierarchical_select_3.v
%Error: tests/gtest/golds/hierarchical_select_3.v:16:5: Unsupported in C: Cell has the same name as variable: 'O'
Baz O (
    ^
        tests/gtest/golds/hierarchical_select_3.v:14:12: ... Location of original declaration
    output O
```

If we want to support this in coreir, we would probably want to run a pass that uniquifies instance/port names before generating verilog

rdaly525:
Ah interesting. I guess that makes sense. Although I think we should still interpret this as valid CoreIR code that should be handled before generating verilog.

CMakeLists.txt
leonardt: Done

leonardt:
Yea, this test basically shows that the internal logic works, it just produced incorrect verilog (but should work as expected if the names are changed in a pass beforehand).

PR #857, opened 2020-03-31
Added pass to inline single instances

rdaly525:
This is a simple pass that greedly inlines any modules which only contains a single instance.

The main purpose of this will be simplifying the clock gating inferring logic, but is also a good way to clean up some of the magma->coreIR compiler steps that do a single level of circuit wrapping.


tests/binary/test_inline_single_instance.py
```diff
@@ -0,0 +1,11 @@
+import delegator
```
rsetaluri: why a python test?

src/passes/transform/inline_single_instances.cpp
```diff
@@ -0,0 +1,23 @@
+#include "coreir.h"
+#include "coreir/passes/transform/inline_single_instances.h"
+
+using namespace std;
+using namespace CoreIR;
```
rsetaluri: avoid this; use `std::` explicitly and wrap this all in `namespace CoreIR { }`

tests/binary/test_inline_single_instance.py
rdaly525: No particular reason other than it being easier to write than bash.

PR #858, opened 2020-03-31
Clock Gating pass

rdaly525:
This PR creates a clock gating pass.

It finds any time there is a mux/register where the output of the register feeds into the input of the mux, and replaces it with a specialized register that has verilog amenable to inferring clock gating. 

Depends on #857


include/coreir/passes/transform/clock_gate.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "coreir.h"
+
+namespace CoreIR {
+
+  namespace Passes {
```
rsetaluri: nit: indent not needed for nested namespaces

src/passes/transform/clock_gate.cpp
```diff
@@ -0,0 +1,123 @@
+#include "coreir/passes/transform/clock_gate.h"
+
+std::string CoreIR::Passes::ClockGate::ID = "clock_gate";
+
+namespace {
+    struct CEInfo {
```
rsetaluri: same as above re: nesting, see below also

src/passes/transform/clock_gate.cpp
```diff
@@ -0,0 +1,123 @@
...
+        CoreIR::Instance* reg = nullptr;
+        CoreIR::Instance* mux = nullptr;
+        int mux_in_port = 0;
+        CEInfo() : can_replace(false) {}
+        CEInfo(CoreIR::Instance* reg, CoreIR::Instance* mux, bool mux_in_port) :
+            can_replace(true), reg(reg), mux(mux), mux_in_port(mux_in_port) {}
+    };
+
+    //Checks if inst of generated module and generated module has the following name
+    bool instance_of(CoreIR::Instance* inst, std::string ns, std::string name) {
```
rsetaluri: these look generic, can we pull them out?

include/coreir/passes/transform/clock_gate.h
rdaly525: Fixed

src/passes/transform/clock_gate.cpp
rdaly525: Fixed

src/passes/transform/clock_gate.cpp
rdaly525: Some of them are definitely generic. I want to make a separate PR for a bunch of these helper APIs

src/passes/transform/clock_gate.cpp
```diff
@@ -0,0 +1,124 @@
...
+namespace {
+struct CEInfo {
+  bool can_replace = false;
+  CoreIR::Instance *reg = nullptr;
+  CoreIR::Instance *mux = nullptr;
+  int mux_in_port = 0;
+  CEInfo() : can_replace(false) {}
+  CEInfo(CoreIR::Instance *reg, CoreIR::Instance *mux, bool mux_in_port)
+      : can_replace(true), reg(reg), mux(mux), mux_in_port(mux_in_port) {}
+};
```
leonardt: Not sure on what's standard (CC @rsetaluri), but any reason this is a struct vs class?

src/passes/transform/clock_gate.cpp
```diff
@@ -0,0 +1,124 @@
...
+// Checks if inst of generated module and generated module has the following
+// name
+bool instance_of(CoreIR::Instance *inst, std::string ns, std::string name) {
+  auto mod = inst->getModuleRef();
+  if (!mod->isGenerated()) {
+    return false;
+  }
+  auto gen = mod->getGenerator();
+  return gen->getName() == name && gen->getNamespace()->getName() == ns;
+}
```
leonardt: This seems like it could be generally useful enough to go into a `transforms/util.{hpp,cpp}`

leonardt: Also (nit), spacing between the function below is inconsistent with others

src/passes/transform/clock_gate.cpp
```diff
@@ -0,0 +1,124 @@
...
+      // not the select first
+      auto not_inst_name = "not_inst" + this->getContext()->getUnique();
+      auto not_inst = def->addInstance(not_inst_name, "corebit.not");
+      def->connect(regSel_driver, not_inst->sel("in"));
+      regSel_driver = not_inst->sel("out");
+    }
+    def->connect(regSel_driver, ce_inst->sel("ce"));
+
+    // Connect the output(s) easiest is to leverage a passthrough
+    auto pt = addPassthrough(info.reg->sel("out"), "_pt");
```
leonardt: Not exactly sure what this does, is it just so that when you remove the other instances, it preserves the connection to the outputs so you can wire up the new register to the passthrough?

src/passes/transform/clock_gate.cpp
rdaly525: I think a struct is fine here since I am just using it as a custom return type for this local function.

src/passes/transform/clock_gate.cpp
rdaly525: Yeah exactly. Since the register output can drive any number of other inputs at any level of its type hierarchy, it is easiest and safest to add a passthrough (check the inlining code for details about the passthrough). I could have done this with the inputs, but since only a single thing can be connected here, the passthrough is a little overkill.

rdaly525:
https://travis-ci.org/github/rdaly525/coreir/builds/669261090 Is passing, but not sure why it isnt reporting back to github

PR #860, opened 2020-04-01
Ce reset

rdaly525:
Updating clock gating logic to work with reg_arst

mantle.hpp: Added regCE_arst for reg_arst and did some code reformatting
clock_gate.cpp: Added code path for reg_arst. Also only do this optimziation when clk and rst are posedge.
*.v, *.json: Added an additional pass of inline_single_instances to clean up the code even more


src/ir/headers/mantle.hpp
```diff
@@ -14,236 +14,315 @@ Namespace* CoreIRLoadHeader_mantle(Context* c) {
...
+      "    if (arst) begin\n"
+      "      value <= init;\n"
+      "    end\n"
+      "    else if (ce) begin\n"
+      "      value <= in;\n"
+      "    end\n"
+      "  end\n"
+      "  assign out = value;";
+    regCEArst->getMetaData()["verilog"] = vjson;
+  }
```
leonardt, line 326: I'm assuming that the added regCEArst is the significant (non-style) change, can you confirm?

src/ir/headers/mantle.hpp
```diff
@@ -14,236 +14,315 @@ Namespace* CoreIRLoadHeader_mantle(Context* c) {
...
+      "input [width-1:0] in",
+      "input ce",
+      "output [width-1:0] out",
+      "input clk",
+      "input arst",
+    };
+    vjson["definition"] =
+      ""
+      "  reg [width-1:0] value;\n"
+      "  always @(posedge clk, posedge arst) begin\n"
```
leonardt, line 316: Do we anticipate a need for negedge reset?

src/ir/headers/mantle.hpp
rdaly525: Yeah for this file it is. Sorry let me add a better description of the changes. Didnt realize the diff was so massive.

src/passes/transform/clock_gate.cpp
```diff
@@ -74,26 +85,40 @@ bool ClockGate::runOnModule(Module *m) {
...
   for (auto ipair : def->getInstances()) {
     auto inst = ipair.second;
-    if (instance_of(inst, "coreir", "reg")) {
+    bool is_coreir_reg = instance_of(inst, "coreir", "reg");
+    bool is_coreir_reg_arst = instance_of(inst, "coreir", "reg_arst");
+    if (is_coreir_reg || is_coreir_reg_arst) {
+      // NYI clock being negedge
+      if (!inst->getModArgs().at("clk_posedge")->get<bool>()) {
+        continue;
+      }
```
leonardt, line 94: If we're not supporting negedge reset yet, we'll need a similar check here.

rdaly525: Fixed

rsetaluri:
@rdaly525 in future, let's try to keep formatting and content PR's sepearate

PR #864, opened 2020-04-08
Use connection mapping for inline values

leonardt:
This augments the inline_verilog feature to accept a mapping of symbols to connections (select paths) that coreir will handle the replacement of.  The goal of this was to have coreir manage the names of the final verilog values, rather than having magma guess them up front.  For this initial implementation, I tested this on the flatten types pass logic by introducing a new object "symbol_table" in the metadata that tracks a mapping from unflattened names to flattened names.  This allows the verilog backend to interpolate the flattened names in the final inline verilog.

In the future we can look into building a more robust symbol table implementation, but for now this logic is so simple that just using the JSON seems reasonable enough.  One option would be to build an API wrapper around the json representation, which has the value of being able to serialize the symbol table for intermediate stages (e.g. run coreir to produce new coreir, with the original symbol table mapping, that could then be used by a future pass of coreir).  An alternative is to have an in-memory data structure for the symbol table (we could then define serialize/deserialize logic for this to achieve the same feature for multiple coreir stages).

The main interface change is that `inline_verilog` metadata is now an object with two fields:
1) "str" corresponds to the inline verilog string that may contain symbols using Python's format string syntax, e.g. `{_special_name_to_be_filled_in_by_coreir}`
2) "connect_references" is an object that maps these symbols to a correspond select path, e.g. `"_special_name_to_be_filled_in_by_coreir": "self.I0.tuple_field.arr.0.2"`.

In the backend, the verilog code generator looks up the connection reference, checks the symbol table for a mapping (following chained mappings, in case multiple passes rename symbols), and replaces the reference name in the verilog string with the final verilog name.

### DIFF
--- a/CMakeLists.txt.in
+++ b/CMakeLists.txt.in
@@ -7,7 +7,7 @@ project(verilogAST-download NONE)
 include(ExternalProject)
 ExternalProject_Add(verilogAST
   GIT_REPOSITORY    https://github.com/leonardt/verilogAST-cpp.git
-  GIT_TAG           inline-verilog-symbol-map
+  GIT_TAG           master
   SOURCE_DIR        "${CMAKE_CURRENT_BINARY_DIR}/verilogAST-src"
   BINARY_DIR        "${CMAKE_CURRENT_BINARY_DIR}/verilogAST-build"
   CONFIGURE_COMMAND ""
--- a/CMakeLists.txt.in
+++ b/CMakeLists.txt.in
@@ -7,7 +7,7 @@ project(verilogAST-download NONE)
 include(ExternalProject)
 ExternalProject_Add(verilogAST
   GIT_REPOSITORY    https://github.com/leonardt/verilogAST-cpp.git
-  GIT_TAG           master
+  GIT_TAG           inline-verilog-symbol-map
   SOURCE_DIR        "${CMAKE_CURRENT_BINARY_DIR}/verilogAST-src"
   BINARY_DIR        "${CMAKE_CURRENT_BINARY_DIR}/verilogAST-build"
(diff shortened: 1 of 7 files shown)


rsetaluri:
Agree that symbol table should be an in memory data structure. If we need to serialize it to json we always can. But it's fine as json for now if that' easiest for keeping it for the lifetime of the module.

Again, It seems that maintaining and updating the symbol table across passes that could modify the symbols is necessary here. I see that the symbol table is created after flatten_types and not touched again. Could this be an issue? What if flatten_types isn't run at all (or is it almost always run?)?

leonardt:
we can check for existence of the symbol table before creating it in flatten types as a temporary solution, the verilog backend assumes flatten types is always run, so a symbol table should exist based on that.

rdaly525:
Couple things: 

First, it is definitely strange to create the symbol table during the flatten types pass. This is very adhoc and I would prefer for it to be an API call or a separate pass that creates a symbol table for all modules which other passes (like flattentypes) can update.

Second, just in general this does not seem safe nor extensible. Lets meet to discuss options for the symbol table. I have some ideas.

rsetaluri:
@leonardt @rdaly525 so were we on the same page re: getting this merged and then doing the right thing soon?

rdaly525:
Yeah, sure. Tracking the inline_verilog cleanup in #867

leonardt:
arg this wasn't quite ready yet, there were still some outstanding bugs with the magma integration, I have some changes staged.  I'll submit another PR to resolve them, but FYI this code won't work with magma master as is.  This should be okay for most people using the pycoreir binary release, but anything using coreir master and inline verilog will be broken at the moment

PR #866, opened 2020-04-10
Round dev

rdaly525:
Changing the default conversion of float -> bfloat to be rounding instead of truncating


src/simulator/interpret.cpp
```diff
@@ -7,6 +7,26 @@ using namespace std;
 
 namespace CoreIR {
 
+  BitVector BfloatRoundToNearestEven(const BitVector& val) {
+    assert(val.bitLength() == 32);
+
+    auto half = BitVector(32, 0x0000800);
```
jeffsetter, line 13: This bitvector should have an additional zero: 
`auto half = BitVector(32, 0x00008000);`

rdaly525: Good catch, fixing

PR #861, opened 2020-04-01
added clang-format

rdaly525:
This is a start for the repo cleanup effort. Eventually I would like to force all C++ files to adhere to this.

PR #869, opened 2020-04-17
Clang format 2

rsetaluri:


rsetaluri:
@rdaly525 compare to #861

.clang-format
```diff
@@ -0,0 +1,126 @@
+---
+Language:        Cpp
+AccessModifierOffset: -1
```
rsetaluri, line 3: like this as -1 vs -2

leonardt:
LGTM, is there a simple way to integrate this into CI (e.g. block if clang-format hasn't been run).  Perhaps we can run clang-format and use git to see if there's any diffs, and set an error code if so

rsetaluri:
Solved by #879

PR #870, opened 2020-04-18
Verilog c api

rsetaluri:


src/coreir-c/coreir-c.cpp
```diff
@@ -108,6 +114,68 @@ extern "C" {
...
+    std::vector<std::string> passes {
+      "rungenerators",
+      "removebulkconnections",
+      "flattentypes",
+      verilog_pass
+    };
+    context->runPasses(passes, namespaces);
+    auto pass = static_cast<Passes::Verilog*>(
+        context->getPassManager()->getAnalysisPass("verilog"));
+    defer(pass->clear());
```
rdaly525, line 155: Why are you deferring this clear function?

rsetaluri: so that it always gets called. otherwise I would have to make sure to have it before every `exit()` point.

rsetaluri: e.g. L167, L174, L177

PR #871, opened 2020-04-20
Hotifx cansel instance

leonardt:
Fixes a bug where canSel fails trying to select an instance (with no subsequent port).  In this case, it should just check whether the instance exists, rather than trying to select an empty select string on the instance.


leonardt:
Also fixes up indents to be consistent

PR #874, opened 2020-04-20
Cast width of binop primitive outputs

leonardt:
Depends on https://github.com/leonardt/verilogAST-cpp/pull/46 (see for related discussion)

### DIFF
--- a/tests/gtest/port_order_golden.v
+++ b/tests/gtest/port_order_golden.v
@@ -4,7 +4,7 @@ module Add8_cin (
     output [7:0] a,
     input CIN
 );
-assign a = (({1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,CIN}) + z) + x;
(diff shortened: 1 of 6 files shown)


tests/gtest/port_order_golden.v
```diff
@@ -4,7 +4,7 @@ module Add8_cin (
     output [7:0] a,
     input CIN
 );
-assign a = (({1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,CIN}) + z) + x;
+assign a = 8'((8'(({1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,1'b0,CIN}) + z)) + x);
```
leonardt, line 7: In the future we can try to clean this up to avoid the nested casts, it's a little tricky because they're only nested after the wire inlining pass, so we'd probably need to run another pass to clean up redundant casts (since at codegen time, each op still gets its own wire, so we need the cast since we can't assume the wire will be removed)

PR #876, opened 2020-04-23
Use RPATH ORIGIN

rsetaluri:


hofstee:
Why did we even need to deal with RPATH in the first place? I know the modifications I made earlier were because if you did `make install` followed by `make clean` then things would break. But what about originally? Seems like it's causing a fair number of the headaches in the build script.

rsetaluri:
at this point, I think things are just a patch work of getting things tot work in various cases. Not sure of the original reason for us, but generally it allows to set the search path at link time.

I'm using origin (search "ORIGIN" in http://man7.org/linux/man-pages/man8/ld.so.8.html) because the search path should be relative to the original build location.

rdaly525:
@hofstee The reason for using rpath is so that coreir can do runtime loading of the coreir libraries like commonlib.

rsetaluri:
Going to use a different approach to fix this...

PR #879, opened 2020-04-24
Clang format final

rdaly525:
This PR updates the entire repo to consistently use a single clang-format style

Also adds a script apply_clang_format.sh to apply the clang format


.clang-format
```diff
@@ -0,0 +1,125 @@
...
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: AfterColon
```
rsetaluri, line 47: Actually feel pretty strongly about this...

```
class Foo {
  Foo(int x) : 
    mem(x),
    other(false) {
...
}
```
looks really weird to me

```
class Foo {
  Foo(int x)
    : mem(x),
      other(false) {
...
}
```
 seems much better.

rsetaluri:
also seems to be failing...

.clang-format
rdaly525: Eh, that looks as weird to me. But I dont really care as long as things always line up on the next indentation level.

rsetaluri: ok, think `BeforeColon` here should do that.

rdaly525: Fixed.

rdaly525:
@rsetaluri, @leonardt I think this should pass CI. Please re-review if desired.

PR #880, opened 2020-04-24
Fix bfloat simulation by using bitvectors

jeffsetter:
After trying an example recently with bfloat value 13.125 * 10, I found that the round to even was not working as intended. By using bitvectors in the arguments, it simulates correctly.


rdaly525:
This makes it seem really strange that it was working at all beforehand... But it seems reasonable to need to compare equality with BitVector types

PR #886, opened 2020-05-01
Resolve extremely occasional inout codegen ordering problem

leonardt:
CoreIR does not assume that connections are deterministically ordered (i.e. conn.first and conn.second do not always match the JSON).  This resulted in the very occasional test failures for the inout code generation logic (which assumed this was always a deterministic relationship).  This adds a fix to make sure the output is deterministic.

I was able to consistently reproduce the issue with 15 test runs of the intermediate connection test: https://github.com/rdaly525/coreir/commit/434163e44f2b3f4878c8326380399d8721928500
verified it resolved the issue with the same test setup: https://github.com/rdaly525/coreir/commit/eec74dca23169e5189e2f338163de5ea9d35ad20
then i removed the repeated tests so our test suite doesn't take too long: https://github.com/rdaly525/coreir/commit/c4542de32bcc6a9263d15b5d3ef150290e7181ea

### DIFF
--- a/include/coreir/ir/common.h
+++ b/include/coreir/ir/common.h
@@ -97,6 +97,9 @@ static std::map<std::string, std::set<std::string>> coreMap({
 
 void mergeValues(Values& v0, Values v1);
 
+// Compare select paths
+bool SPComp(const SelectPath& l, const SelectPath& r);
+
 }  // namespace CoreIR
 
 #endif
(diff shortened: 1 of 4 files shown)


leonardt:
Here's the example failure: https://travis-ci.org/github/rdaly525/coreir/jobs/682147080

PR #889, opened 2020-05-02
Isolate primitives

rdaly525:
This pass is a sort of "reverse inlining" pass. It essentially collects all instances of coreir primitives and isolates them into a single instance of their own module. This is very useful when needing to operate on only the coreir instances of the module.

Also, I could imagine this being a first step in being able to generate always_comb/always_ff blocks during verilog generation.

### DIFF
--- a/src/passes/transform/isolate_primitives.cpp
+++ b/src/passes/transform/isolate_primitives.cpp
@@ -69,6 +69,7 @@ bool IsolatePrimitives::runOnModule(Module* m) {
 
   // External connections
   for (auto const& [wP, _, portName] : boundary_info) {
+    UNUSED(_);
     auto portSP = SelectPath({"self", portName});
     pdef->connect(wP->getSelectPath(), portSP);
   }
@@ -77,11 +78,13 @@ bool IsolatePrimitives::runOnModule(Module* m) {
   // Create instance of prim module and connect appropriately
   auto primInst = def->addInstance("primitive_" + c->getUnique(), primModule);
   for (auto const& [_, connE, portName] : boundary_info) {
+    UNUSED(_);
     def->connect(connE, primInst->sel(portName));
   }
 
   // remove all primtiive instnaces
   for (auto const& [_, inst] : primitiveInstances) {
+    UNUSED(_);
     def->removeInstance(inst);
   }
 
--- a/src/passes/transform/isolate_primitives.cpp
+++ b/src/passes/transform/isolate_primitives.cpp
@@ -1,0 +1,84 @@
+#include "coreir/passes/transform/isolate_primitives.h"
+
+std::string CoreIR::Passes::IsolatePrimitives::ID = "isolate_primitives";
+
+// Creates a separate module that isolates all the primitive (coreir/corebit) instances
+
+namespace CoreIR {
+namespace Passes {
+bool IsolatePrimitives::runOnModule(Module* m) {
+  if (!m->hasDef()) {return false;}
+  auto def = m->getDef();
+  auto c = this->getContext();
+
+  //get a map of only primitive instances
+  auto primitiveInstances = filterOver(
+    def->getInstances(),
+    [](auto it) {
+      auto nsname = it.second->getModuleRef()->getNamespace()->getName();
+      return nsname == "coreir" || nsname == "corebit";
+    }
+  );
+
+  //early out
+  if (primitiveInstances.size()==0) {return false;}
+
+  std::vector<std::pair<Wireable*, Wireable*>> internal_connections; //prim <=> prim
+  std::vector<std::pair<Wireable*, Wireable*>> external_connections; //prim <=> other
+  for (auto const& [wA, wB] : def->getConnections()) {
+    //Put into two piles external and internal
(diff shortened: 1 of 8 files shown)


src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
+#include "coreir/passes/transform/isolate_primitives.h"
+
+std::string CoreIR::Passes::IsolatePrimitives::ID = "isolate_primitives";
```
rsetaluri, line 3: can just put this inside the namespace

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+// Creates a separate module that isolates all the primitive (coreir/corebit) instances
+
+namespace CoreIR {
+namespace Passes {
+bool IsolatePrimitives::runOnModule(Module* m) {
```
rsetaluri, line 9: nit: newline before

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (!m->hasDef()) {return false;}
```
rsetaluri, line 10:
```suggestion
  if (!m->hasDef()) return false;
```

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+  auto def = m->getDef();
+  auto c = this->getContext();
+
+  //get a map of only primitive instances
```
rsetaluri, line 14:
```suggestion
  // Get a map of only primitive instances.
```

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+  auto primitiveInstances = filterOver(
+    def->getInstances(),
+    [](auto it) {
+      auto nsname = it.second->getModuleRef()->getNamespace()->getName();
+      return nsname == "coreir" || nsname == "corebit";
+    }
+  );
+
+  //early out
+  if (primitiveInstances.size()==0) {return false;}
```
rsetaluri, line 24:
```suggestion
  if (primitiveInstances.size()==0) return false;    // early out
```

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  std::vector<std::pair<Wireable*, Wireable*>> internal_connections; //prim <=> prim
+  std::vector<std::pair<Wireable*, Wireable*>> external_connections; //prim <=> other
+  for (auto const& [wA, wB] : def->getConnections()) {
+    //Put into two piles external and internal
```
rsetaluri, line 29:
```suggestion
  std::vector<std::pair<Wireable*, Wireable*>> internal_connections;  // prim <=> prim
  std::vector<std::pair<Wireable*, Wireable*>> external_connections;  // prim <=> other
  for (auto const& [wA, wB] : def->getConnections()) {
    // Put into two piles external and internal.
```

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+    bool AIsPrimitive = primitiveInstances.count(wA->getSelectPath()[0]);
+    bool BIsPrimitive = primitiveInstances.count(wB->getSelectPath()[0]);
+    if (AIsPrimitive && BIsPrimitive) {internal_connections.push_back(std::make_pair(wA, wB));}
+    else if (AIsPrimitive) {external_connections.push_back(std::make_pair(wA, wB));}
+    else if (BIsPrimitive) {external_connections.push_back(std::make_pair(wB, wA));}
+  }
+
+  //Construct module type interface
```
rsetaluri, line 37: Standalone comments should be `// start of comment` and comments at the end of line should be like `my_code_here();  // start of comment`.

I think running clang-format will fix all of these.

src/passes/transform/isolate_primitives.cpp
```diff
@@ -0,0 +1,84 @@
...
+  std::vector<std::tuple<Wireable*, Wireable*, std::string>> boundary_info;
+  for (auto const& [wP, wE] : external_connections) {
+    std::string portName = c->getUnique();
+    ports.push_back(std::make_pair(portName, wP->getType()));
+    boundary_info.push_back(std::make_tuple(wP, wE, portName));
+    def->disconnect(wP, wE);
+  }
+
+  //Create new prim module and definition
+  auto primModule = c->getNamespace("_")->newModuleDecl(m->getName() + "_primitives", c->Record(ports));
```
rsetaluri, line 50: is this naming safe? is there a safer, more robust way to do this naming?

src/passes/transform/isolate_primitives.cpp
rdaly525: This is consistent with all the other passes. A PR to change all of them would be better.

src/passes/transform/isolate_primitives.cpp
rdaly525: It is relatively safe since my "_" namespace is basically an internal namespace. But a more systematic way to come up with unique names should be though about.

PR #891, opened 2020-05-11
Change reg to have clear priority over enable to match mantle

jeffsetter:
Follows this discussion: https://github.com/phanrahan/mantle/issues/149

And fixes the CoreIR version of the mantle reg to have clear priority over the enable signal.


jeffsetter:
Actually CoreIR does it correctly already... woops

PR #892, opened 2020-05-27
Add initial support for slices

leonardt:
The main motivation for this feature addition is to improve compiler performance.  There's places in the magma code where we generate large bit-blasted connections for slices and we're hoping that by representing slices explicitly we can reduce the number of API calls (as well as internal logic that recurses over the bit-blasted select/connections).  It should also improve readability of the code by collapsing the bit-blasted slices in the json and verilog.

This basic feature addition simply adds the notion of ":" in the select path to stand for slices.  I'm open to alternatives, particularly if there's any issues with this syntax you forsee.  I added minimal code to support this feature, which basically involved modifying canSel and sel to support the slice selstr, as well as the verilog backend to code generate the slice.

### DIFF
--- a/include/coreir/ir/common.h
+++ b/include/coreir/ir/common.h
@@ -19,7 +19,7 @@ bool isNumber(std::string s);
 bool isPower2(uint n);
 
 bool isSlice(std::string selstr);
-void parseSlice(std::string selstr, uint* low, uint* high);
+std::pair<int, int> parseSlice(const std::string& selstr);
 
 // Used to make sure string formats are valid for inst names, module names, etc
 void checkStringSyntax(std::string& str);
--- a/include/coreir/ir/common.h
+++ b/include/coreir/ir/common.h
@@ -18,6 +18,9 @@ typedef std::set<Connection, ConnectionCompFast> ConnectionsFast;
 bool isNumber(std::string s);
 bool isPower2(uint n);
 
+bool isSlice(std::string selstr);
+void parseSlice(std::string selstr, uint* low, uint* high);
+
 // Used to make sure string formats are valid for inst names, module names, etc
 void checkStringSyntax(std::string& str);
 
--- a/src/ir/common.cpp
+++ b/src/ir/common.cpp
@@ -36,7 +36,7 @@ bool isSlice(std::string selstr) {
     isNumber(selstr.substr(colon_pos + 1));
 }
 
-std::pair<int, int> void parseSlice(const std::string& selstr) {
+std::pair<int, int> parseSlice(const std::string& selstr) {
   // Assumes guarded by isSlice and is a valid slice.
   size_t colon_pos;  // will be set by stoi to first non-num (colon)
   int low = std::stoi(selstr, &colon_pos, 0);
--- a/src/ir/common.cpp
+++ b/src/ir/common.cpp
@@ -26,7 +26,22 @@ namespace CoreIR {
 bool isNumber(string s) {
   return !s.empty() && s.find_first_not_of("0123456789") == string::npos;
 }
+
 bool isPower2(uint n) { return (n & (n - 1)) == 0; }
+
+bool isSlice(std::string selstr) {
+  auto colon_pos = selstr.find_first_of(":");
+  if (colon_pos == std::string::npos) { return false; }
+  return isNumber(selstr.substr(0, colon_pos)) &&
+    isNumber(selstr.substr(colon_pos + 1));
+}
+
+void parseSlice(std::string selstr, uint* low, uint* high) {
+  // assumes guarded by isSlice and is a valid slice
+  size_t colon_pos;  // will be set by stoi to first non-num (colon)
+  *low = std::stoi(selstr, &colon_pos, 0);
+  *high = std::stoi(selstr.substr(colon_pos + 1), nullptr, 0);
+}
 
 bool SPComp(const SelectPath& l, const SelectPath& r) {
   string ls = toString(l);
--- a/src/ir/common.cpp
+++ b/src/ir/common.cpp
@@ -31,16 +31,17 @@ bool isPower2(uint n) { return (n & (n - 1)) == 0; }
 
 bool isSlice(std::string selstr) {
   auto colon_pos = selstr.find_first_of(":");
-  if (colon_pos == std::string::npos) { return false; }
+  if (colon_pos == std::string::npos) return false;
   return isNumber(selstr.substr(0, colon_pos)) &&
     isNumber(selstr.substr(colon_pos + 1));
 }
 
-void parseSlice(std::string selstr, uint* low, uint* high) {
-  // assumes guarded by isSlice and is a valid slice
+std::pair<int, int> void parseSlice(const std::string& selstr) {
+  // Assumes guarded by isSlice and is a valid slice.
   size_t colon_pos;  // will be set by stoi to first non-num (colon)
-  *low = std::stoi(selstr, &colon_pos, 0);
-  *high = std::stoi(selstr.substr(colon_pos + 1), nullptr, 0);
+  int low = std::stoi(selstr, &colon_pos, 0);
+  int high = std::stoi(selstr.substr(colon_pos + 1), nullptr, 0);
+  return {low, high};
 }
 
 bool SPComp(const SelectPath& l, const SelectPath& r) {
--- a/src/ir/types.cpp
+++ b/src/ir/types.cpp
@@ -37,6 +37,10 @@ bool Type::isBaseType() {
   return isa<BitType>(this) || isa<BitInType>(this) || isa<BitInOutType>(this);
 }
 
+bool sliceIsValid(uint low, uint high, ArrayType* arr) {
+  return (low >= 0) && (high <= arr->getLen()) && (low < high);
+}
+
 Type* Type::sel(string selstr) {
   if (auto rt = dyn_cast<RecordType>(this)) {
     ASSERT(rt->getRecord().count(selstr), "Bad Select!");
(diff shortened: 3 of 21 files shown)


leonardt:
An issue to consider:

What happens when we select/connect an array port that's already been connected in a slice? I'm not sure if/where coreir does this checking for normal array selects, but we should be sure to be consistent here and to catch this error.  Fortunately for us, I don't know if this is a major user facing issue, since at the magma level we already do this checking, so at least in the magma->coreir flow we don't necessarily need this error checking except for validating the compiler internals (although this safety check would be nice for future development that might introduce bugs of this nature).

include/coreir/ir/common.h
```diff
@@ -18,6 +18,9 @@ typedef std::set<Connection, ConnectionCompFast> ConnectionsFast;
 bool isNumber(std::string s);
 bool isPower2(uint n);
 
+bool isSlice(std::string selstr);
+void parseSlice(std::string selstr, uint* low, uint* high);
```
rsetaluri, line 22:
```suggestion
std::pair<int, int> parseSlice(const std::string& selstr);
```

 (may need to `#include <pair>` at the top)

src/ir/common.cpp
```diff
@@ -26,8 +26,23 @@ namespace CoreIR {
 bool isNumber(string s) {
   return !s.empty() && s.find_first_not_of("0123456789") == string::npos;
 }
+
 bool isPower2(uint n) { return (n & (n - 1)) == 0; }
 
+bool isSlice(std::string selstr) {
+  auto colon_pos = selstr.find_first_of(":");
+  if (colon_pos == std::string::npos) { return false; }
```
rsetaluri, line 34:
```suggestion
  if (colon_pos == std::string::npos) return false;
```

src/ir/common.cpp
```diff
@@ -26,8 +26,23 @@ namespace CoreIR {
...
+  return isNumber(selstr.substr(0, colon_pos)) &&
+    isNumber(selstr.substr(colon_pos + 1));
+}
+
+void parseSlice(std::string selstr, uint* low, uint* high) {
+  // assumes guarded by isSlice and is a valid slice
+  size_t colon_pos;  // will be set by stoi to first non-num (colon)
+  *low = std::stoi(selstr, &colon_pos, 0);
+  *high = std::stoi(selstr.substr(colon_pos + 1), nullptr, 0);
+}
```
rsetaluri, line 44:
```suggestion
std::pair<int, int> void parseSlice(const std::string& selstr) {
  // Assumes guarded by isSlice and is a valid slice.
  size_t colon_pos;  // will be set by stoi to first non-num (colon)
  int low = std::stoi(selstr, &colon_pos, 0);
  int high = std::stoi(selstr.substr(colon_pos + 1), nullptr, 0);
  return {low, high};
}
```

src/ir/types.cpp
```diff
@@ -37,6 +37,10 @@ bool Type::isBaseType() {
   return isa<BitType>(this) || isa<BitInType>(this) || isa<BitInOutType>(this);
 }
 
+bool sliceIsValid(uint low, uint high, ArrayType* arr) {
```
rsetaluri, line 40:
```suggestion
bool sliceIsValid(uint low, uint high, ArrayType* const arr) {
```

src/ir/types.cpp
```diff
@@ -45,7 +49,15 @@ Type* Type::sel(string selstr) {
     return rt->getRecord().at(selstr);
   }
   else if (auto at = dyn_cast<ArrayType>(this)) {
-    ASSERT(isNumber(selstr), selstr + " needs to be a number!");
+    if (isSlice(selstr)) {
+      uint low, high;
+      parseSlice(selstr, &low, &high);
```
rsetaluri, line 54:
```suggestion
      int low;
      int high;
      std::tie(low, high) = parseSlice(selstr);
```

src/ir/types.cpp
```diff
@@ -70,7 +82,12 @@ bool Type::canSel(string selstr) {
     return rt->getRecord().count(selstr);
   }
   else if (auto at = dyn_cast<ArrayType>(this)) {
-    if (!isNumber(selstr)) return false;
+    if (!isNumber(selstr) && !isSlice(selstr)) return false;
+    if (isSlice(selstr)) {
+      uint low, high;
+      parseSlice(selstr, &low, &high);
```
rsetaluri, line 88:
```suggestion
      int low;
      int high;
      std::tie(low, high) = parseSlice(selstr);
```

src/ir/wireable.cpp
```diff
@@ -3,6 +3,7 @@
 #include "coreir/ir/common.h"
 #include "coreir/ir/context.h"
 #include "coreir/ir/error.h"
+
```
rsetaluri, line 6: nit: unnec. new line

src/passes/analysis/verilog.cpp
```diff
@@ -698,6 +705,22 @@ convert_to_verilog_connection(Wireable* value, bool _inline) {
...
     }
+    else if (isSlice(item)) {
+      ASSERT(
+        i == select_path.size() - 1,
+        "Assumed flattened types have array slice as last element "
+        "in select path");
+      if (std::holds_alternative<std::unique_ptr<vAST::Identifier>>(
+            curr_node)) {
+        uint low, high;
+        parseSlice(item, &low, &high);
```
rsetaluri, line 716:
```suggestion
        int low;
        int high;
        std::tie(low, high) = parseSlice(item);
```

leonardt:
Added some logic to verifyconnectivity to handle the slices (this was needed for magma support), for some reason I can't reproduce the magma issue for a test case, so working on that.  Also, there might be a more clever (performant) way to handle skipping the indices, but this should work.

rdaly525:
Looking through this now. I think I understand from a convenience point of view, but why not just leverage coreir.slice instances?

rdaly525:
> An issue to consider:
> 
> What happens when we select/connect an array port that's already been connected in a slice? I'm not sure if/where coreir does this checking for normal array selects, but we should be sure to be consistent here and to catch this error. Fortunately for us, I don't know if this is a major user facing issue, since at the magma level we already do this checking, so at least in the magma->coreir flow we don't necessarily need this error checking except for validating the compiler internals (although this safety check would be nice for future development that might introduce bugs of this nature).

This is done in the ModuleDef.validate method.

rdaly525:
Some additional tests in order to add this feature: Does a slice work on any array type (or just arrays of bits)? Does slices of slices work correctly? Does selects of slices work correctly?

leonardt:
@rdaly525 the main reason is performance, we want to avoid any extra logic in handling slices, so supporting them natively seems to make the most sense (since it's not too complicated) versus piggy backing on other mechanisms (which incurs more API calls to represent the same thing)

leonardt:
For now I'm restricting slices only to work on array of bits, I think more general slices are harder to handle so let's start with the simpler case first (also will let us validate what kind of performance increases we can see).  I can add a validator check to make sure we catch the error when slices are used otherwise.  The slice of slices and select of slices tests are good, I'll look into them (but FWIW, magma won't use either of these so we could also make these errors for now)

rdaly525:
Okay, as long as you catch and label those errors as NYI.

rsetaluri:
re: arrays of things other than bits...it's not clear to me why this isn't supported automatically? is it a matter of the connectivity validation being harder?

leonardt:
@rsetaluri hmm, I haven't though about it, I can take a look into how hard it would be to support.  my initial thought was that since nested arrays are being flattened out anyways, we wouldn't be code generating these nested slices, so their representation is less helpful for our performance issue.  In fact, I think this shows one place where complication might occur.  When flattening a nested array, we would need to flatten the corresponding slices, which would be doable but likely a bit complex.

rsetaluri:
Yeah that would need some extra logic - makes sense. Hopefully moving to N-D arrays in verilog will help.

rdaly525:
Yeah a couple of passes might interact with this feature: 

FlattenTypes
RemoveBulkConnections
PackBitConstants
PackConnections
UnpackConnections

Although really this illuminates the need for better testing on these already written passes.

leonardt:
* FlattenTypes - is this problematic if we don't support general nested arrays? (Just arrays of bits) I can look into a test to see if it's a problem (perhaps the flattened connected wireables don't get updated properly, but I think my basic test uses nested arrays and it works fine).
* RemoveBulkConnections - I can update this to basically remove slices in a similar fashion to bulk connections, which should match the expected behavior
* PackBitConstants - this goes from 1-bit consts to a packed const, not sure if this is problematic unless we want the packed const to use a slice connection? Or is it if the consts are wired up to a slice?
* PackConnections, UnpackConnections - I'm not exactly sure what these do, are they related to remove bulk connections?

rdaly525:
I think for a lot of these, just making sure the addPassthrough() function works, means that these should work.

> * FlattenTypes - is this problematic if we don't support general nested arrays? (Just arrays of bits) I can look into a test to see if it's a problem (perhaps the flattened connected wireables don't get updated properly, but I think my basic test uses nested arrays and it works fine).

I suspect this will not need any updates till nd-array support

> * RemoveBulkConnections - I can update this to basically remove slices in a similar fashion to bulk connections, which should match the expected behavior

This pass basically just transforms any circuit to that where connections are only bits or bitvectors. So, you should keep slices of bitvectors, but simplify any more complicated slices.

> * PackBitConstants - this goes from 1-bit consts to a packed const, not sure if this is problematic unless we want the packed const to use a slice connection? Or is it if the consts are wired up to a slice?

Seems like no changes required.

> * PackConnections, UnpackConnections - I'm not exactly sure what these do, are they related to remove bulk connections?

I do not recall the exact semantics of what these do... Ill update in a few minutes

rdaly525:
Other methods that come to mind that could be affected:
Wireable.disconnectAll
Wireable.getAllSelects
Wireable.getLocalConnections

leonardt:
I reviewed that code and I think because this just piggy backs on the existing connection/select logic (it doesn't introduce any new code paths there, it just adds a new path that can introduce a new select/connection that behaves just like others), so I think those should work as expected since there are no changes to the visible API for selects/connections.

Is there a pass I can use to test the use of these API calls? Maybe inlining a slice would be a good test

leonardt:
It seems like I can use createcombview for the second two, and instancegraph or sanitize names for the first one, I'll try them out

leonardt:
ok looks like disconnect all is used in detach field (instance graph node) which is used inside flatten types, so the basic test involves wiring slices inside a nested type that is flattened so that seems to work as expected.

leonardt:
Tried out comb view with slices connected to/from a register and seems to work as expected

PR #893, opened 2020-05-29
Improve logic for flattening arrays

leonardt:
This improves the performance of flattening arrays by first flattening the first child, then copying the result of flattening for all the children.

Before, we were iteratively flattening each child, but since each child contains the same type, they can share the logic for flattening.

A related improvement that I'll look into is the memoization of flattening types.  Ideally we can store a pointer -> flattened port list map and just fetch that when we see a type multiple times.  @rdaly525 are types cached in the sense that we'll be able to do a simple pointer value hashing scheme for this?


leonardt:
Let's hold off on merging this, my initial profile seems to indicate little performance benefit or a small regression,  but I'll do some more digging (perhaps there's a smarter way to avoid the copies for the select path, which is probably a bulk of the performance cost).

PR #894, opened 2020-06-05
MuxN -> If

leonardt:
Depends on https://github.com/leonardt/verilogAST-cpp/pull/49

This augments the inlining of commonlib.muxn instances to if/elseif/else statements inside an always *.  We also update the declaration to use a reg instead of a wire (in my experience, synthesis tools will complain if you try to assign to a wire inside an always * block).

We should consider making muxn a primitive since we're giving it some special behavior in the backend, but the current implementation doesn't require any changes to the core so it's less invasive while being useful.  This is guarded behind the `inline` flag so it won't be enabled in default flows.

I also moved the AST transformers used for inlining into a separate file for organization


rdaly525:
This is great progress for improving the verilog code quality. 

I do not want to have a MuxN as a primitive though, as it has no real correspondence with the SMT QF BitVector primitives and introduces specialization creep into the IR. Instead, I think we should move to an approach of having multiple different kinds of ModuleDef/GeneratorDef instead of ones just for CoreIR implementations. A MuxN would thus just have a particular generator implementation that would generate a verilog AST module containing the always block. Verilog Instances in verilogAST could then be inlined if desired.

Issue #895, opened 2020-06-07
slicing breaks inlining


rdaly525:
@leonardt the new slicing syntax breaks the current inlining code and any code that uses passthroughs.(https://github.com/rdaly525/coreir/blob/master/src/ir/inline.cpp). In particular the case when there are non-matching slices for a bitvector port in the module instantiation and the module definition.

Unfortunately this is quite a complicated fix that requires instantiating coreir.wire nodes when this case arises.

A few fixes in the interim:
1) Disallow inlining when there are slices
2) Always translate slices to single bit selects before inlining

Let me know if this makes sense.

leonardt:
Do you have a test to reproduce the issue?

rdaly525:
@leonardt creating a simple example now

rdaly525:
```
{"top":"global.A",
  "namespaces": {
    "global": {
      "modules": {
        "A": {
          "type": ["Record",[
            ["in", ["Array",16,"BitIn"]],
            ["out", ["Array",16,"Bit"]]
          ]],
            "instances": {
            "i0": {"modref": "global.B"}
          },
          "connections": [
            ["self.in.0:3", "i0.in.0:3"],
            ["self.in.3:16", "i0.in.3:16"],
            ["i0.out", "self.out"]
          ]
        },
        "B": {
          "type": ["Record",[
            ["in", ["Array",16,"BitIn"]],
            ["out", ["Array",16,"Bit"]]
          ]],
          "instances": {
            "n1": {
              "genref": "coreir.neg",
              "genargs": {"width":["Int", 16]}
            }
          },
          "connections": [
            ["self.in.0:10","n1.in.0:10"],
            ["self.in.10:16","n1.in.10:16"],
            ["n1.out","self.out"]
          ]
        }
      }
    }
  }
}
```

running flatten on this:
`>coreir -i example.json -p flatten
`

will produce:
```
{"top":"global.A",
"namespaces":{
  "global":{
    "modules":{
      "A":{
        "type":["Record",[
          ["in",["Array",16,"BitIn"]],
          ["out",["Array",16,"Bit"]]
        ]],
        "instances":{
          "i0$n1":{
            "genref":"coreir.neg",
            "genargs":{"width":["Int",16]}
          }
        },
        "connections":[
          ["self.out","i0$n1.out"]
        ]
      },
      "B":{
        "type":["Record",[
          ["in",["Array",16,"BitIn"]],
          ["out",["Array",16,"Bit"]]
        ]],
[...]

PR #897, opened 2020-06-09
Add fix for inlining slices

leonardt:
See https://github.com/rdaly525/coreir/issues/895 for more context

This adds a step in the inlining process to insert intermediate wires for slice selects so the current inlining implementation works.  There might be alternative solutions to this, but hopefully this fixes the functionality bug with minimal changes to the code.  I played around with some other ideas but it seems pretty complicated to get this right in the actual inlining logic (need to map from indices of one slice to indices of another).  I think this gets us a baseline working version with the least amount of code, but may introduce some overhead in the output (extra wire instances), but I think this might be better than blasting the slices into individual bits.

### DIFF
--- a/src/ir/inline.cpp
+++ b/src/ir/inline.cpp
@@ -1,4 +1,5 @@
 #include "coreir/ir/casting/casting.h"
+#include "coreir/ir/common.h"
 #include "coreir/ir/context.h"
 #include "coreir/ir/generator.h"
 #include "coreir/ir/moduledef.h"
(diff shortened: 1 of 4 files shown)


rdaly525:
Unfortunately it is not just inline that is affected, but any code that uses the "addPassthrough" method. Grep-ing through the source code it looks like at least 8 passes use this function, so we need to a more general solution. Also, if addPassthrough is safe for slices, then inlining is safe for slices.

Similar to your code, we can do a similar tactic of adding coreir.wire during the PTTraverse function (used by addPassthrough). Basically in the first line of that function, when the other wireable is a slice, stamp out a coreir.wire instance.

Your code does an additional optimization where it does not add "unneeded" coreir.wire instances. This should be easy to detect and cleanup in a separate pass. In fact that pass would be analogous to [copy propagation](https://en.wikipedia.org/wiki/Copy_propagation) and would be generally useful as a normal optimization pass.

Let me know if this makes sense. Apologies for not being clearer in the issue.

PR #900, opened 2020-06-17
Support verilog codegen of mem primitive init param

leonardt:
To implement the `init` parameter in the verilog codegen for the mem primitive, I found that changing the parameter type to be a flat bit vector (containing the packed bits of size depth*width for the initial values) to be simpler.  To support verilator, we need to do this in the Verilog code generation since it doesn't support a 2d array as a parameter argument.  More generally, using a generic JSON type for initialization isn't quite safe, since really what we want is a 2d array of size width * depth.  With a generic JSON list, we could treat it as a 2d list of bit values, or a 1d list of ints.  The problem with ints is that we don't have a size for the internal value (so we will have trouble packing them into the flat verilog representation for verilator).  We could possibly carry around this information in the verilog code generation, but with all these complications, I felt it would be easier to maintain if we just used the flat array of bits representation for the initial value.  

With this, we're able to implement the init logic fairly simply with minimal code changes by using a generate block inside verilog.  I've tested this in verilator for the magma primitive (will check in a functional simulation test into the magma repo) and it works, so I assume the commercial simulators should have no problem supporting it.

I'd be open to the idea of "parsing" the JSON version of the init parameter as a list of bit vectors, but then we'd need to use an ad-hoc schema for the contents of the list (which is the reason I decided to just use the flat bit representation instead, since I don't think it's too much trouble for the user to convert to this form, and we avoid the code complexity of parsing the JSON)

### DIFF
--- a/tests/gtest/golds/rom.v
+++ b/tests/gtest/golds/rom.v
@@ -35,7 +35,7 @@ module Memory (
     input CLK
 );
 coreir_mem #(
-    .init({5'11,5'21,5'0,5'5}),
+    .init({5'd11,5'd21,5'd0,5'd5}),
     .depth(4),
     .has_init(1'b1),
     .width(5)
--- a/tests/gtest/golds/rom.v
+++ b/tests/gtest/golds/rom.v
@@ -1,0 +1,63 @@
+module coreir_mem #(
+    parameter has_init = 1'b0,
+    parameter depth = 1,
+    parameter width = 1
+) (
+    input clk,
+    input [width-1:0] wdata,
+    input [$clog2(depth)-1:0] waddr,
+    input wen,
+    output [width-1:0] rdata,
+    input [$clog2(depth)-1:0] raddr
+);
+  reg [width-1:0] data [depth-1:0];
+  parameter [width*depth-1:0] init = 0;
+  generate if (has_init) begin
+    genvar j;
+    for (j = 0; j < depth; j = j + 1) begin
+      initial begin
+        data[j] = init[(j+1)*width-1:j*width];
+      end
+    end
+  end
+  endgenerate
+  always @(posedge clk) begin
+    if (wen) begin
+      data[waddr] <= wdata;
+    end
+  end
+  assign rdata = data[raddr];
+endmodule
+
+module Memory (
+    input [1:0] RADDR,
+    output [4:0] RDATA,
+    input CLK
+);
+coreir_mem #(
+    .init(20'h04d47),
+    .depth(4),
+    .has_init(1'b1),
+    .width(5)
+) coreir_mem4x5_inst0 (
+    .clk(CLK),
+    .wdata(5'h00),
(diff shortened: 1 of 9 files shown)


leonardt:
Hmm, just found this https://github.com/rdaly525/coreir/pull/555/files, looks like this was how things used to be, @rdaly525 is there a reason why we switched to JSON? Do you know if anyone is actively using the memory/rom interface with JSON initialization? (Since this will break the API)

leonardt:
Hmm, the issue may have to do with a huge bit vector for initialization of a large memory?

rdaly525:
This is being used currently by @jeffsetter (and Lake maybe?) and the json was implemented to simplify the interface for specifying the array contents without defining a new Array CoreIR ValueType. I agree that it is an ad hoc json schema and therefore is not the nicest for being part of a coreir primitive. I would be fine with changing it to be more robust.

jeffsetter:
Yes, the current initialization of memories is using a json of initial values. I'd be open to changing the interface if it maintains usability and readability.

leonardt:
Hmm, okay for now I think I might preserve the current interface just to avoid breaking changes.

My main concern will be the representation of initial values great than what the json parsing library can support (I'm assuming they're reading integer literals into a standard C++ type so it's going to be bounded by that representation).  So, I'll propose for now we keep using a list of ints as the initialization, gather the information on the bit width from the memory generator parameters, and throw an exception if we encounter an initialization width great than the supported integer width (I'm guessing it's 32 or 64 wide).  I'm hoping that our current use cases aren't using memory entries wider than that, and we can adds support for arbitrary width init values later on (probably by using some alternative representation for each entry, such as a bit vector string)

leonardt:
For the JSON interface, it looks like you use `init_list[0].get<int>();`, so the user passes the type, so we'll be limited by our widest C++ data type.

leonardt:
Latest version depends on https://github.com/leonardt/verilogAST-cpp/pull/50 and preserves JSON array interface

rdaly525:
I actually did not realize we were using ints. We should be using a json array of bitvectors to fix the arbitrary size issue.

There might be some utility for directly encoding the schema or an value indicating which schema is being used.

tests/gtest/golds/rom.v
```diff
@@ -0,0 +1,63 @@
...
+) (
+    input clk,
+    input [width-1:0] wdata,
+    input [$clog2(depth)-1:0] waddr,
+    input wen,
+    output [width-1:0] rdata,
+    input [$clog2(depth)-1:0] raddr
+);
+  reg [width-1:0] data [depth-1:0];
+  parameter [width*depth-1:0] init = 0;
```
rdaly525, line 14: Interesting way to do this. I think this would work well.

leonardt:
I'll leave that as a separate follow-up PR when needed, since that change would require breaking @jeffsetter 's code, so I'd prefer to not do that until we have to.

PR #904, opened 2020-06-25
Add disable width cast API arg

leonardt:
For https://github.com/rdaly525/coreir/pull/898


rsetaluri:
Glad this API is getting use and "working" :)

PR #907, opened 2020-06-25
Expose disable width cast option in binary

leonardt:
For https://github.com/phanrahan/magma/issues/708


leonardt:
Bump, this is needed by https://github.com/phanrahan/magma/pull/741

leonardt:
Bump, needed for https://github.com/phanrahan/magma/pull/741

leonardt:
Agreed, a simpler temporary solution would be to just prefix them with "v" (e.g. -vw).  I think the issue is that we need to passthrough the arguments to the verilog pass, but that pass is called implicitly by the choice of the ".v" output.  Another option is to provide some argument like "output_pass_args" that will forward arguments to the final pass.

rdaly525:
I like the "output_pass_args" route except I would want to call it something like "serialization_args" or even "codegen_args"

PR #912, opened 2020-06-30
Fix mem leaks

leonardt:
These came up when debugging https://github.com/rdaly525/coreir/pull/911

Minor changes to cleanup some code in the binary exit logic.  Also one style change caught by clang-format

### DIFF
--- a/src/binary/coreir.cpp
+++ b/src/binary/coreir.cpp
@@ -161,7 +161,7 @@ int main(int argc, char* argv[]) {
           outExt == "v",
         "Cannot support out extention: " + outExt);
       if (!split_files) {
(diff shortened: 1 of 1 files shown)


src/binary/coreir.cpp
```diff
@@ -267,5 +269,8 @@ int main(int argc, char* argv[]) {
   }
   LOG(DEBUG) << "Modified?: " << (modified ? "Yes" : "No");
 
+  if (delete_sout) delete sout;
+  delete c;
```
rsetaluri, line 273: Prefer: `deleteContext(c);`

(Not a huge fan of this pattern anyway, but since it's used elsewhere might as well keep it standard)

PR #913, opened 2020-06-30
Remove static ID logic

leonardt:
Redo of https://github.com/rdaly525/coreir/pull/844, see that thread for some of the design issues.  I mainly did the minimal changes required to avoid using the static ID logic so we can distribute the static binary without the double free problems.  There might be a better way to design this pattern without using static, but this should at least avoid the double free error messages.

The tutorial files were also updated, but they also include style changes since I think we didn't run clang-format on them when we made the style change (my editor automatically formatted them).


rdaly525:
@leonardt Looks like the CI reporting is possibly stuck in a bad state. Might want to retrigger.

PR #920, opened 2020-07-16
Use longname for verilog module codegen

leonardt:
Fixes https://github.com/phanrahan/magma/pull/772

### DIFF
--- a/tests/gtest/golds/slice_combview.v
+++ b/tests/gtest/golds/slice_combview.v
@@ -30,11 +30,11 @@ mantle_wire__typeBit16 _$_U2 (
     .in(value_src_out),
     .out(_$_U2_out)
 );
-coreir_reg__width16_snk value_snk (
+coreir_coreir_reg__width16_snk value_snk (
     .clk(CLK),
     .in(_$_U1_in)
 );
-coreir_reg__width16_src value_src (
+coreir_coreir_reg__width16_src value_src (
     .out(value_src_out)
 );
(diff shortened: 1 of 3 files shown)


tests/gtest/golds/slice_combview.v
```diff
@@ -30,11 +30,11 @@ mantle_wire__typeBit16 _$_U2 (
     .in(value_src_out),
     .out(_$_U2_out)
 );
-coreir_reg__width16_snk value_snk (
+coreir_coreir_reg__width16_snk value_snk (
     .clk(CLK),
     .in(_$_U1_in)
 );
-coreir_reg__width16_src value_src (
+coreir_coreir_reg__width16_src value_src (
```
leonardt, line 37: This is a little weird, but I think expected? Basically, when comb view serializes, it creates new modules for the src/snk in the coreir namespace, but I think it includes the longname in the new module name.  Regardless, the serialization of comb view to verilog was just to test the flattening slice feature, so I think this is fine (these modules don't have definitions, since they're just used to represent the source/sink in the comb view graph, so can't verify that they match the definition name since it doesn't exist)

leonardt: Confirmed it uses longname, so the double coreir_ prefix is expected, see https://github.com/rdaly525/coreir/blob/df962c4ebd6be2f3af70c24d8b39b54c24e645e7/src/passes/transform/transform2combview.cpp#L94

PR #921, opened 2020-07-17
Added wire passes. Fixed isolate pass

rdaly525:


include/coreir/passes/common.h
```diff
@@ -84,7 +84,9 @@ void initializePasses(PassManager& pm) {
   pm.addPass(new Passes::RemoveUnconnected());
   pm.addPass(new Passes::RemoveSingleMuxes());
   pm.addPass(
-    new Passes::WireClocks("wireclocks-coreir", c->Named("coreir.clkIn")));
+    new Passes::WireClocks("wireclocks-clk", c->Named("coreir.clkIn"), "CLK"));
+  pm.addPass(
+    new Passes::WireClocks("wireclocks-arst", c->Named("coreir.arstIn"), "ARST"));
```
leonardt, line 89: Should the names here be lowercase? It's "clk" down below before this change

rdaly525: It does not really matter since this is just the created port name. This pass searches based off of type and not string name. I chose upper case to be more consistent with magma sequential generated code.

PR #922, opened 2020-07-20
Adds support for magma's user defined namespace

leonardt:
Magma added support for a user_namespace, but currently magma only
compiles the namespace containing the top module.  This doesn't work
when we want some modules in the user_namespace, but other modules in
the global namespace (e.g. imported verilog modules that shouldn't have the
user_namespace prefix).  These changes are required so that magma can
compile the whole context instead of just the namespace of the top
module.

* Add Context.setTop to C API (so magma can set the top module in a
  context)
* Add saveToFile of context to C API (so magma can save an entire
  context instead of just the namespace of the top module)
* Add option to saveToFile to skip default imported library namespaces
  like mantle and commonlib.  These don't need to be in the JSON (and
  break downstream gold files when switching to this flow), skipping
  them improves readability (less noise).
* Adds logic in json pass to skip a namespace if it's empty (e.g. global
  namespace may be empty when magma is using only the user_namespace)

### DIFF
--- a/src/coreir-c/coreir-c.cpp
+++ b/src/coreir-c/coreir-c.cpp
@@ -124,6 +124,10 @@ bool COREContextRunPasses(
   return context->runPasses(vec_passes, vec_namespaces);
 }
 
+void COREContextSetTop(COREContext* context, COREModule* module) {
+  rcast<Context*>(context)->setTop(rcast<Module*>(module));
+}
+
 bool CORECompileToVerilog(
   COREContext* ctx,
   COREModule* top,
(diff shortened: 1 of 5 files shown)


src/coreir-c/coreir-c.cpp
```diff
@@ -156,10 +160,11 @@ bool CORECompileToVerilog(
...
   std::vector<std::string> namespaces{"global"};
-  std::vector<std::string> passes{"rungenerators",
-                                  "removebulkconnections",
-                                  "flattentypes",
-                                  verilog_pass};
+  std::vector<std::string> passes{
+    "rungenerators",
+    "removebulkconnections",
+    "flattentypes",
+    verilog_pass};
```
leonardt, line 167: Unrelated style change from auto clang format

PR #925, opened 2020-07-27
Adds flattentypes ndarray arg

leonardt:
This preserves multi-dimensional arrays of bit/bits in the flattening
pass (so we can code generate multi-dimensional arrays in verilog).  I
was originally going to call this "flattentuples", but it's not quite
that because if we have an array of tuples, we do the old flattening
process, so there are still some arrays that are flattened.  I think one
way to handle this might be to have a pass before this that transforms
arrays of tuples into tuples of arrays, so for now I'll check this
version in since this will get us started with the common case
(multi-dimensional array of bits) and leave the multi-dimensional
array of tuple case for later.


leonardt:
Mainly for backwards compatibility, I didn't want to cause a breaking change for existing code.  I could see about making this an option with it off by default, but I think the original flatten types behavior should be preserved and maybe gracefully deprecated.

PR #926, opened 2020-07-29
Only skip mantle wire when inline is enabled

leonardt:


leonardt:
Fixes https://travis-ci.org/github/phanrahan/magma/builds/712777451

PR #927, opened 2020-07-29
Allow test_isolate to run without install

leonardt:
This updates the commonlib loading logic so that it works without having
corier installed into a system path (the current logic assumes it's
available somewhere, but sometimes when you're developing locally and
you don't have a copy installed into the system)


rsetaluri:
Does this work if the commonlib.so is dynamically linked? consulting @rdaly525 for any weirdnesses here...

leonardt:
Yes, this was how libraries were designed to be loaded (see https://github.com/rdaly525/coreir/blob/master/doc/LibraryGuide.md#libraries), it's also how the other tests all load the libraries (https://github.com/rdaly525/coreir/blob/master/tests/gtest/test_verilog.cpp#L166-L168)

Issue #932, opened 2020-08-09
Segfault in verilog generation for camera pipeline


dillonhuff:
@rdaly525 I'm generating coreir for the cgra implementation of camera pipeline (Jeff gave me basic versions of the compute units), but when I try to create verilog for the coreir implementation to test it before PnR I get an error:

```bash
dhuff@kiwi:~/clockwork$ ${COREIR_PATH}/bin/coreir --load_libs commonlib --input camera_pipeline.json --output camera_pipeline.v
Aborted (core dumped)
```

The error backtrace is:

```bash
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff65868b1 in __GI_abort () at abort.c:79
#2  0x00007ffff799bb75 in CoreIR::convert_value(CoreIR::Value*) () from /home/dhuff/clockwork/lib/libcoreir.so
#3  0x00007ffff79a2c0d in CoreIR::compile_module_body(CoreIR::RecordType*, CoreIR::ModuleDef*, bool, std::set<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&) () from /home/dhuff/clockwork/lib/libcoreir.so
#4  0x00007ffff79a5991 in CoreIR::Passes::Verilog::compileModule(CoreIR::Module*) ()
   from /home/dhuff/clockwork/lib/libcoreir.so
#5  0x00007ffff79a6515 in CoreIR::Passes::Verilog::runOnInstanceGraphNode(CoreIR::InstanceGraphNode&) ()
   from /home/dhuff/clockwork/lib/libcoreir.so
#6  0x00007ffff7879d91 in CoreIR::PassManager::runInstanceGraphPass(CoreIR::Pass*) ()
   from /home/dhuff/clockwork/lib/libcoreir.so
#7  0x00007ffff787a1c0 in CoreIR::PassManager::runPass(CoreIR::Pass*, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&) () from /home/dhuff/clockwork/lib/libcoreir.so
[...]

PR #933, opened 2020-08-10
Add support for rom2 instancing coreir.mem with init

rsetaluri:
See #932 

TODO: add test for rom2


rsetaluri:
Thanks. Want to make sure the verilog is good for Dillon then I will merge.

rsetaluri:
It seems like a general pass like flatten_types to convert all parameters into "POD"s would be a way to do it. Of course what a "POD" is may depend on the context (verilog vs sv).

dillonhuff:
@rdaly525 @rsetaluri can we merge this? I have down_sample, cascade, pointwise, gaussian, and strided conv working in coreir verilog simulation, and I'd like to add camera pipeline to that list.

PR #936, opened 2020-08-11
[verilog] Simplify instance input connection logic

leonardt:
This improves the input connection logic for instance statements to use
an existing identifier if it's the sole driver of an input.  This avoids
having to emit an input wire for all instance inputs (this was leading
to a code size explosion compared to the old logic which would just
directly insert the driver's identifier).  We avoid the inlining issue
by marking the driver as blacklisted from inlining (the original reason
we introduced these inputs wires was so we could mark them as
blacklisted so expressions weren't inlined into the instance statement
for tool compatability, but now we can just mark the driver instead of
emitting a new wire).

This should reduce the code size explosion, since now we'll only
introduce an input wire if we encounter a non-identifier driver (e.g. a
concat for a non-bulk connection, or assignments to the input wire for
elements of a unpacked array).  This will be particularly good for the
non-inline case where all we were doing was introducing extra wires
(since inlining wasn't being run, the inline expressions issue isn't a
problem).

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -984,6 +984,14 @@ void assign_inouts(
   };
 }
 
+bool isMemModule(Module* module) {
+  return module->isGenerated() &&
+    ((module->getGenerator()->getName() == "mem" &&
+      module->getGenerator()->getNamespace()->getName() == "coreir") ||
+     (module->getGenerator()->getName() == "rom2" &&
+      module->getGenerator()->getNamespace()->getName() == "memory"));
+}
+
 // Traverses the instance map and creates a vector of module instantiations
 // using connection_map to wire up instance ports
 std::vector<std::variant<
@@ -1135,13 +1143,7 @@ Passes::Verilog::compileModuleBody(
         }
       }
     }
-    bool is_mem_inst = instance_module->isGenerated() &&
-      ((instance_module->getGenerator()->getName() == "mem" &&
-        instance_module->getGenerator()->getNamespace()->getName() ==
-          "coreir") ||
-       (instance_module->getGenerator()->getName() == "rom2" &&
-        instance_module->getGenerator()->getNamespace()->getName() ==
-          "memory"));
+    bool is_mem_inst = isMemModule(instance_module);
     // Handle module arguments
     if (instance.second->hasModArgs()) {
       for (auto parameter : instance.second->getModArgs()) {
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -1105,15 +1105,25 @@ Passes::Verilog::compileModuleBody(
             wire_name,
             field_type,
             entries);
-        auto id_ptr = dynamic_cast<vAST::Identifier*>(driver.get());
+        auto target_id = dynamic_cast<vAST::Identifier*>(driver.get());
         if (
-          id_ptr &&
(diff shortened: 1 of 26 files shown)


src/passes/analysis/verilog.cpp
```diff
@@ -1116,18 +1105,36 @@ Passes::Verilog::compileModuleBody(
...
-        ((instance_module->getGenerator()->getName() == "mem" &&
-          instance_module->getGenerator()->getNamespace()->getName() == "coreir") ||
-         (instance_module->getGenerator()->getName() == "rom2" &&
-          instance_module->getGenerator()->getNamespace()->getName() == "memory"));
+      ((instance_module->getGenerator()->getName() == "mem" &&
+        instance_module->getGenerator()->getNamespace()->getName() ==
+          "coreir") ||
+       (instance_module->getGenerator()->getName() == "rom2" &&
+        instance_module->getGenerator()->getNamespace()->getName() ==
+          "memory"));
```
leonardt, line 1137: Unrelated auto style change

src/passes/analysis/verilog.cpp
```diff
@@ -1088,6 +1072,11 @@ Passes::Verilog::compileModuleBody(
             instance_name + "." + field,
             _inline);
         if (concat->unpacked && !is_inlined) {
+          this->makeDecl(wire_name, field_type, body, false);
+          this->wires.insert(wire_name);
+          verilog_connections->insert(
+            field,
+            std::make_unique<vAST::Identifier>(wire_name));
           wireUnpackecDriver(body, std::move(concat), vAST::make_id(wire_name));
```
rsetaluri: not relevant but see a typo here 😮

src/passes/analysis/verilog.cpp
rsetaluri: This is getting pretty big...can we put this in a helper function like `is_memory_primitive()`

src/passes/analysis/verilog.cpp
```diff
@@ -939,7 +923,8 @@ void assign_module_outputs(
           _inline);
       if (concat->unpacked) {
         wireUnpackecDriver(body, std::move(concat), vAST::make_id(field));
-      } else {
+      }
```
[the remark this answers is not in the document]
leonardt: I run automatic clang-format to do the style related stuff and this is what it does, so we should change the .clang-format if we want something different

src/passes/analysis/verilog.cpp
leonardt: unrelated to the functionality of this change, but sure

src/passes/analysis/verilog.cpp
leonardt: Fixed in earlier commit

src/passes/analysis/verilog.cpp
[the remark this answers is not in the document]
leonardt: https://github.com/rdaly525/coreir/blob/master/.clang-format#L37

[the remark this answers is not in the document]
leonardt: From: http://clang.llvm.org/docs/ClangFormatStyleOptions.html

image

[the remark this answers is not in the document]
rsetaluri: Ahh thanks for clarifying. Think maybe Ross preferred that and I forgot.

PR #937, opened 2020-08-11
[verilog] omit input wire for Bit <-> Bits[1]

leonardt:
While we are technically generating correct code, this omits the input
wire if we're connecting a bits[1] to a bit since in verilog that's
allowed and that allows us to omit another wire (our previous change
still causes an out of memory error in the garnet test, so the hope is
that removing this wire will help us get under the memory usage limit)


leonardt:
I do have a copy of the old verilog generated pre ndarray (since I was using it diff for debugging an issue), this commit brings it back below the original file size (in line count), so this should at least rule out file size as the problem now.

PR #943, opened 2020-08-18
Cache getTopParent logic

leonardt:
This shaves 5-7 seconds off the garnet compile time (17ish vs 22ish) by cacheing the getTopParent logic.  This function came up in the profile because I believe when we're dealing with low-level selects, we're repeatedly traversing the select path to get to the top parent.  @rdaly525 I believe this is safe since with selects, the parent shouldn't ever change.  This way, if we call getTopParent on two array children, they should share a cached parent rather than recomputing the select path.

Also a minor related change in verilog.cpp where we call getTopParent() twice, the compiler may already be optimizing this out, but just in case we reuse the result of the first invocation.

### DIFF
--- a/include/coreir/ir/wireable.h
+++ b/include/coreir/ir/wireable.h
@@ -16,6 +16,7 @@ class Wireable : public MetaData {
   WireableKind kind;
   ModuleDef* container;  // ModuleDef which it is contained in
   Type* type;
+  Wireable* topParent = NULL;
 
   std::set<Wireable*> connected;
 
--- a/include/coreir/ir/wireable.h
+++ b/include/coreir/ir/wireable.h
@@ -16,7 +16,7 @@ class Wireable : public MetaData {
   WireableKind kind;
   ModuleDef* container;  // ModuleDef which it is contained in
   Type* type;
-  Wireable* topParent = NULL;
+  Wireable* topParent = nullptr;
 
   std::set<Wireable*> connected;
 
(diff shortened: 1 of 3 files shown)


leonardt:
A further optimization on top of this might be to store the parent logic when we're constructing the selects (so we don't have to recompute it and cache it), but this should work until we have that change.

include/coreir/ir/wireable.h
```diff
@@ -16,6 +16,7 @@ class Wireable : public MetaData {
   WireableKind kind;
   ModuleDef* container;  // ModuleDef which it is contained in
   Type* type;
+  Wireable* topParent = NULL;
```
rsetaluri, line 19:
```suggestion
  Wireable* topParent = nullptr;
```

src/ir/wireable.cpp
```diff
@@ -199,16 +199,20 @@ LocalConnections Wireable::getLocalConnections() {
 }
 
 Wireable* Wireable::getTopParent() {
-  Wireable* top = this;
```
rsetaluri, line 202:
```c++
  if (topParent != nullptr) return topParent;
  if (isa<Select>(this) || isa<InstanceSelect>(this)) {
    Select* s = isa<Select>(this) ?
        cast<Select>(this) :
        cast<InstanceSelect>(this);
    topParent = s->getParent()->getTopParent();
  }
  else {
    topParent = this;
  }
  return topParent;
```

PR #950, opened 2020-08-25
Fpops

rdaly525:


jeffsetter:
One last operator I would suggest is "mux"

rdaly525:
@jeffsetter mux looks to have already been defined.

jeffsetter:
Ah I missed that it had its own line. List looks good to me

PR #965, opened 2020-10-02
Add capability to exclude module from inline_single_instances using metadata

leonardt:
Intended usecase is for garnet, where `inline_single_instances` is
changing a module name (by inlining it) which is breaking the PD flow
(hardcoded constraints for the outside module).  This adds a metadata
flag we can use to disable inline_single_instances from being run on a
module.


rdaly525 (approved):
Good feature!

rsetaluri (approved):

PR #968, opened 2020-10-09
Add SyncReadMemory primitive

leonardt:
This adds a SyncReadMemory primitive that is similar to coreir.mem with
the sync_read parameter enabled, except that it also supports a read
enable port.

The motivation is that we want to generate the `ren` logic in the
verilog using the `if` inside the `always` block for synthesis.
However, we can't smash this inside the existing memory generator since
it would require an "optional" `ren` port that isn't used when sync_read
is disabled.  It doesn't seem like verilog supports this optional port
pattern.

One alternative to defining a separate primitive would be to always
generate a `ren` port in the verilog and ignore it if `sync_read` is
disabled.

Ideally we'd have this in place of the sync_read parameter, but I left
the old logic in for backwards compatability.  With sync_read, if the
user doesn't want to use the enable logic, we can just set the input to
1 and presumably the downstream tool can optimize that out.

Setting ren to a constant 1 when the enable logic is not desired seemed
like a better pattern than having a dangling `ren` port when there's a
combinational read, but the downside is we have to use a separate
primitive.

Perhaps an alternative long-term solution would be to add a
`sync_read_enable` argument to the memory and have it generate different
verilog code based on the parameter.  This might be a good effort for
other cases, e.g. where we use `generate if` for the verilog memory init
logic (instead we could just generate the `init` code if the parameter
is enabled), but this requires quite a bit of work since the verilog
primitive code generation wasn't designed to support this (right now we
just store a static verilog string to emit, rather than generating the
string dynamically, we'd need to create some API for this).

### DIFF
--- a/include/coreir/definitions/coreVerilog.hpp
+++ b/include/coreir/definitions/coreVerilog.hpp
@@ -777,6 +777,7 @@ void CoreIRLoadVerilog_coreir(Context* c) {
       "    if (ren) rdata_reg <= data[raddr];\n"
       "  end\n"
       "  assign rdata = rdata_reg;\n";
-    core->getGenerator("sync_read_mem")->getMetaData()["verilog"] = vjson;
+    Namespace* memory = c->getNamespace("memory");
+    memory->getGenerator("sync_read_mem")->getMetaData()["verilog"] = vjson;
   }
 }
--- a/include/coreir/definitions/coreVerilog.hpp
+++ b/include/coreir/definitions/coreVerilog.hpp
@@ -1,4 +1,5 @@
 #include "verilogAST.hpp"
+#include "coreir/definitions/memoryVerilog.hpp"
 
 namespace vAST = verilogAST;
 
@@ -519,18 +520,7 @@ void CoreIRLoadVerilog_coreir(Context* c) {
         "input [width-1:0] in",
         "output [width-1:0] out",
       },
-    },
-    {
-      "mem",
-      {
-        "input clk",
-        "input [width-1:0] wdata",
-        "input [$clog2(depth)-1:0] waddr",
-        "input wen",
-        "output [width-1:0] rdata",
-        "input [$clog2(depth)-1:0] raddr",
-      },
-    },
+    }
   };
 
   Namespace* core = c->getNamespace("coreir");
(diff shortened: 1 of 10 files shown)


leonardt:
Also, FWIW, this could be easily implemented as a wrapper around memory (register with enable on the read port), but we lose the ability to generate the desired verilog pattern

rsetaluri:
Thanks for the summary of issues.

I think this general problem is interesting and definitely one of interest to @rdaly525 re: linking and generators. Which is basically related to the problems of optional ports and specializing generator implementations based on bound gen. params.

For now, I think this is the on the right side of the trade-off. But a few things I noted:

* Is there anyway to make logic at https://github.com/rdaly525/coreir/pull/968/files#diff-6c4fb4c3cc23c2c09bda419439646981R401-R404 more robust? That stuff seems to be repeated and has to change everytime we introduce a new memory like thing.
* How much of the actual verilog code in this generator is duplicated from the original memory? One way to ease the tension of creating an entirely new generator would be to be able to reuse most of the actual verilog string. Perhaps abstracting this away into a routine that takes in meta-generator parameters and produces the verilog string could get some more reuse.

leonardt:
> Is there anyway to make logic at https://github.com/rdaly525/coreir/pull/968/files#diff-6c4fb4c3cc23c2c09bda419439646981R401-R404 more robust? That stuff seems to be repeated and has to change everytime we introduce a new memory like thing.

This relates to the verilog backend's special handling of the memory init parameter.  Basically, we parse the JSON argument into a format that is valid verilog (in general we don't support JSON parameters yet).  We could at least abstract these checks into helper functions to organize the code better (`is_module_with_mem_init`).  I don't think there's a way around pattern matching on memories unless we added explicit support in the C++ type hierarchy (then we could check if it's some instance of an abstract Memory class), but since this is represented by the coreir objects, we need to check the name for this.

> How much of the actual verilog code in this generator is duplicated from the original memory? One way to ease the tension of creating an entirely new generator would be to be able to reuse most of the actual verilog string. Perhaps abstracting this away into a routine that takes in meta-generator parameters and produces the verilog string could get some more reuse.

Most of it except for the `if (ren) <do update>` line, I'll try refactoring that into a function that generates the static strings, that's a good idea of at least avoiding maintaining the common code in two places.

rdaly525:
I echo Raj's comments, and also I would strongly prefer if we had this in the memory namespace instead of the coreir namespace.

Ideally we can translate the raw verilog string code at some point to a generator function that takes in the CoreIR generator parameters and returns a sound verilogAST representation of the generated module.

leonardt:
> Ideally we can translate the raw verilog string code at some point to a generator function that takes in the CoreIR generator parameters and returns a sound verilogAST representation of the generated module.

I think the biggest barrier to this is having something like a verilog -> verilogAST parser which is non-trivial.  Manually constructing the AST nodes for something like the memory is quite tedious, but perhaps less work than writing a verilog parser (the up-side is that the subset of verilog that verilogAST uses is quite simple so less work than a full verilog parser).  Alternatively, we could look into a simple custom syntax for constructing the nodes (something less verbose than the C++ code), or perhaps some convenience macros could get us pretty far

rdaly525 (approved):

PR #969, opened 2020-10-15
Duplicate c string before returning in type-c

makaimann:
This is a (possible) fix for a compiler error on an untested compiler:
```
clang++ --version
Apple clang version 12.0.0 (clang-1200.0.26.2)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
```

The current master fails to build with the following error:
```
[ 80%] Building CXX object src/simulator/CMakeFiles/coreirsim.dir/print_c.cpp.o
/Users/makaim/repos/forks/makaimann/coreir/src/coreir-c/types-c.cpp:74:10: error: returning address of local temporary object [-Werror,-Wreturn-stack-address]
  return rcast<NamedType*>(namedType)->toString().c_str();
```

This PR would just call `strdup` to duplicate the string before returning. This means that memory needs to be freed externally. To be honest, I'm not sure this is the best fix. I'd be happy to make changes based on feedback.


leonardt:
O.O I'm surprised this hasn't caused a problem, but we probably use/discard the string before calling any other functions that use the stack reference.

Requiring the user to free memory is not the worst, I think we could have the Python API wrap the string pointer in an object and free the memory when it's garbage collected.  However, tracing through this it uses some dynamic logic to compute the name:
* https://github.com/rdaly525/coreir/blob/2af836c85f05173bb5f12b59f6684d5579515f92/include/coreir/ir/types.h#L107
* https://github.com/rdaly525/coreir/blob/2af836c85f05173bb5f12b59f6684d5579515f92/src/ir/globalvalue.cpp#L16

If, instead, we stored the string for the lifetime of the C++ object (and had the object manage the memory), then we could just return a pointer to that string and not worry about managing it.

leonardt:
CC @rdaly525 do you think we could update the logic mentioned above to be computed during object initialization rather than on demand?

rdaly525:
Good bug catch! Yeah, we can precompute it.

leonardt:
Hmm, looked into this some more, it seems that `toString` is a virtual method implemented by other things like `Module`, where I'm not sure pre-computing can be done (we can't compute the the string at initialization time since the module may change over it's lifetime).  The alternative is to cache toString calls in side the object, but again this would require us to support the fact that the module may change and need to track when the string is invalidated which seems like a lot of complexity.  I'm inclined to think that the solution presented here might be the best, since it is a C API we can require that the user manage the memory of the string.

leonardt:
@rsetaluri any ideas on how we might avoid the user having to manage the allocated memory?

rsetaluri:
A fairly non-intrusive and memory-safe option would be to transfer fixed size (say 256 bytes) buffers instead of char pointers. Of course if a name takes > 256 chars this wouldn't work but I think this is a fairly common pattern for these ffi endpoints. If the strings we're transferring look more like serializations of objects then we should probably have a more sophisticated mechanism (e.g. protobuf).

Without making any other changes, this local change to copy the string is probably not so bad. It definitively introduces a memory leak but given there's probably other ones in the code base, I'd argue a few bytes isn't so bad.

rsetaluri:
2 other options actually:

1. CoreIR::Context keeps a scratchpad of random stuff that it should deallocate upon destruction. It can even be associated with a specific module or namespace or whatever, so that scratchpad memory is deallocated upon the specific objects destruction. This is also not very intrusive and is pretty robust actually. The only question is if this is really different than just allowing a memory leak since the context probably lives for the entirety of the program.

2. We could allocate memory on the client side and pass in a mutable pointer to pointer rather than returning a pointer. I'm not sure how this interacts with `ctypes` but conceptually is probably the most "right" thing to do since the client is really the one that owns that string (and it's memory).

leonardt:
2. might be the most idiomatic C solution for the C-API, so I'm leaning towards that (the user allocates the memory so is explicitly responsible for freeing it).  The problem here is that we'd need to know the length of the string so for all these string style methods we'd need to add something like `getStrLen` so we can allocate a buffer, then use a method to copy the string into the buffer.  Alternatively we could use some sort of `snprintf` style interface where we specify the size of the buffer and check the return value to see if we need to increase the size.

Both aren't great since option (a) requires us to duplicate the string generation (once to get the length, once to fill the buffer), or implement cacheing with invalidation (then we might as well use the option where the C++ object stores the string memory, but then there's a chance that a pointer may be come "invalid" if the string changes, depending if the user expects it to change as the entity updates).  (b) requires us to guess the buffer size, which may mean occasionally allocating too much space, and other times having to reallocate/reinvoke.

the `strdup` option requires the user be responsible for managing the memory, but maybe that's not so bad since that's how the `strdup` interface itself works (so this style of having a library method allocate memory for you isn't unheard of)

rsetaluri:
For the issue of getting string length, we could make all string return values of the C API be wrapped in a lightweight std::string like data structure. Like 

```C++
struct SizedString {
    char* c_str;
    int len;
};
```

And then we always transfer these things around. On the C++ side we allocate memory for `c_str` then return a `SizedString` object which can be properly deallocated on the client side. I think this works.

leonardt:
Yea that seems like a reasonable pattern, I can look into an implementation since pycoreir will need to be updated.  Fortunately, I think we can just consume this structure, copy it into a Python string (since we need to use `.decode()` anyways) and then discard the memory.  I wonder if auditing similar code with memory management can help us free up some utilization where we are either (a) having the context manage the memory (which means it'll live for the lifetime of the process, basically a leak) or (b) expecting the user to free the memory but it's not being freed (an actual leak).  If this is using a non-trivial amount of memory, some of our performance problems might be related (wasting time in the heap allocator).  I can take a look at this too.

rsetaluri:
@makaimann @leonardt would suggest that if this is at all blocking, to merge the memory leak and refactor later

makaimann:
It's not a blocker for me. But it doesn't compile on Mac OS Catalina, so it might be nice to merge it. Totally up to you all though.

rsetaluri:
Fixed by #982.

PR #973, opened 2020-10-27
changed ieee=True for CW

rdaly525:


rsetaluri (approved):
looks good, thanks for add -> mul name change

PR #976, opened 2020-11-03
Pass through ieee_compliance param to instance

rsetaluri:


rsetaluri:
@rdaly525 i think this got left behind when making the change before (#973)

rdaly525 (approved):

PR #977, opened 2020-11-10
Add undriven off lint flag for verilator compat

leonardt:
Similar to the comments we use for the `term` primitives to silence the
verilator unused warnings, this adds comments to silence verilator
undriven warnings when using the `undriven` primitie.


leonardt:
The changes in verilator_compat.json ensure this is covered by the test that consumes that file, produces verilog, and runs verilator to check if there are any lint errors 9https://github.com/rdaly525/coreir/blob/d34406cda663ca93db1f1dea66b64bbbb1431858/tests/binary/test_issues.py#L4)

rdaly525 (approved):

PR #978, opened 2020-12-07
Add suport for .sv extension

leonardt:
Needed for https://github.com/phanrahan/magma/issues/880

Allows the user to choose `.sv` as an output.  In the future, we should
consider forcing the `disable-ndarray` option when `.v` is used (i.e.
choosing `.sv` enables this feature since it should be compatible with
tools that support `.sv`).   However, for backwards compatability, I
think we should leave the behavior as is (or else various downstream
users will have different code generated by default since everyone is
using `.v` right now but some are expected the ndarray feature enabled
for `.v`).

### DIFF
--- a/include/coreir/passes/analysis/verilog.h
+++ b/include/coreir/passes/analysis/verilog.h
@@ -100,7 +100,8 @@ class Verilog : public InstanceGraphPass {
   void writeToStream(std::ostream& os) override;
   void writeToFiles(
     const std::string& dir,
-    std::unique_ptr<std::string> product_file);
+    std::unique_ptr<std::string> product_file,
+    std::string outExt);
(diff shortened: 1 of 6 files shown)


rsetaluri (approved):
Agree we should clean up the logic as far as ndarray/system-verilog is concerned, but this seems fine for now.

leonardt:
When we update magma to choose .sv versus .v depending on whether ndarray is enabled, I think we can make the change here to disable ndarray for .v

rdaly525 (approved):

rdaly525:
Looks like there is some call that is missing the outExt parameter causing the failing CI. Alternatively you could have a default value.

PR #982, opened 2021-02-04
Fix CORENamedTypeToString c-API

rsetaluri:
This change adds a context-managed scratch-pad memory which can be used
safely, without risk of a memory leak. The c-API CORENamedTypeToString
previously returned a pointer to a temporary, which is unsafe. Instead,
we allocate context-managed memory and return a buffer contained there.

See https://github.com/rdaly525/coreir/pull/969.


rdaly525 (approved):
Seems like a more generic mechanism for the rest of the C-API allocations scattered throughout the context.

PR #985, opened 2021-02-23
Update C-API for record to use deterministic order

leonardt:
Possible fix for
https://github.com/phanrahan/magma/pull/920#discussion_r580838423

Before, this was iterating over a std::map which is not ordered.  This
changes it to use the `getFields` API which should return the same order
for the fields.


rdaly525 (approved):
Good catch. This kind of issue could easily exist in other portions of the code. Ideally at some point we move over to an insertion-ordered map for this record representation.

rsetaluri (approved):

PR #986, opened 2021-02-24
Add compile guard support

leonardt:


tests/gtest/test_verilog.cpp
```diff
@@ -345,6 +345,26 @@ TEST(VerilogTests, TestInlineIndex) {
   deleteContext(c);
 }
 
+TEST(VerilogTests, TestCompileGuard) {
```
rsetaluri, line 348: can we add a simpler test? this has a bunch of inline verilog stuff. i think a simple test with just an instance wrapped in compile guard would be useful.

rsetaluri (approved):
request for simpler test. o/w looks good

PR #988, opened 2021-02-27
Add support for inverted compile guard

leonardt:
Support for the `ifndef` pattern has been requested.


leonardt:
Updated this to use type=defined/undefined

rsetaluri (approved):

Issue #990, opened 2021-03-01
[feature request] adding prefix to coreir/commonlib/... verilog modules


leonardt:
The desired use case is generating a design where all the relevant verilog modules have a specific prefix (this is used to distinguish the design components from others when integrated in a larger design with blocks from multiple sources).  Right now we can use the namespace feature to insert a prefix before user defined modules, but we can't insert this prefix for coreir libraries (e.g. primitives, commonlib).  Mechanically this should be easy enough to support in the backend, but how can we expose this as an option to the user? One simple option might be to have some metadata field (e.g. verilog prefix), then at the magma level we could insert these for all the non user namespace modules referenced.

rsetaluri:
I think the best option is to have an **unsafe** `output_name` field on any CoreIR module (maybe instance too), which will override the default CoreIR name generation logic. Throwing a bunch of logic into name generation @ CoreIR level seems complicated and hard to get right. We could do logic at the higher levels (e.g. magma) to make sure that any `output_name` overrides are safe.

leonardt:
A seemingly simple option: provide some option for verilog code generation for a prefix that goes before all modules (e.g. `coreir --verilog_module_name_prefix=foo`).  Then, simply do a pass on all the verilogAST modules and insert the prefix.  The main benefit here is that this should work for code that includes multiple namespaces (e.g. we may want to have separate namespaces to handle shared names, but at the end of the day we want everything to have a shared prefix for the above use case).  This feels simpler than having to do a pass at the magma level to set output_name for all modules.   Also, I'm not sure that would work for commonlib muxes for example, since in magma we reference them as instances of generators, to the actual generated module does not have a handle until rungenerators is invoked later on (so we'd need to do this logic at the coreir level not the magma level)

rsetaluri:
A code-generation/verilogAST specific pass seems fine to me. I was concerned if it was a CoreIR pass/abstraction, it would be clunky but doing that in verilogAST seems reasonable.

rdaly525:
The verilogAST pass seems reasonable to me.

PR #991, opened 2021-03-03
[#990] Add compile option for verilog prefix

leonardt:


src/binary/coreir.cpp
```diff
@@ -49,6 +49,9 @@ int main(int argc, char* argv[]) {
     "n,namespaces",
     "namespaces to output: '<namespace1>,<namespace2>,<namespace3>,...'",
     cxxopts::value<std::string>()->default_value("global"))(
+    "verilog-prefix",
```
rsetaluri, line 52: tbh I like dash instead of underscore for cmd line args, but seems like the standard is underscores

```suggestion
    "verilog_prefix",
```

src/passes/analysis/verilog.cpp
```diff
@@ -1594,14 +1600,60 @@ bool Passes::Verilog::runOnInstanceGraphNode(InstanceGraphNode& node) {
...
+  virtual std::unique_ptr<vAST::ModuleInstantiation> visit(
+    std::unique_ptr<vAST::ModuleInstantiation> node) {
+    if (renamed_modules.count(node->module_name)) {
+      node->module_name = prefix + node->module_name;
+    }
+    return node;
+  }
+};
+
 void Passes::Verilog::writeToStream(std::ostream& os) {
```
rsetaluri, line 1636: why do this in `writeToStream`? I think we should do this in the main run function.

src/passes/analysis/verilog.cpp
```diff
@@ -1594,14 +1600,60 @@ bool Passes::Verilog::runOnInstanceGraphNode(InstanceGraphNode& node) {
   return false;
 }
 
+bool add_prefix(
```
rsetaluri, line 1603: is there any reason to do this as a pass rather than at code generation time? I like this more as a code-generation option than a pass.

one of the reasons for this is that it makes symbol-tracking easier, if it's done "on site" rather than as a post-pass.

leonardt: It's complicated to do this at code generation time, since there are a lot of different places where names are computed (e.g. different logic for the primitives versus standard modules versus modules coming from verilog strings).  It's doable, but doing it after the standard names are created is much simpler.  Could we just update the symbol table at this step since we can just map the non-prefixed names to the new prefix names?

src/passes/analysis/verilog.cpp
leonardt: Which run function are you referring to? We can't do this on runOnInstanceGraphNode since we need to do a "global" pass to see which modules are renamed since not all are, so we need to do this before or after the traversal.  I chose to do it here since at this point, the modules are all generated and we are simply traversing them and inserting the prefix before the final step in code generation, which seemed like a reasonable place to do it.

src/binary/coreir.cpp
leonardt: Standard as in general practice or standard is in coreir? I think using dash is standard in general practice, there are some coreir options that don't use it but I think those should be changed, rather than changing this.

From https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
> Option names are typically one to three words long, with hyphens to separate words.

rsetaluri: agreed, dash is *standard* standard (and better), just meant as far as the CoreIR repo is concerned. We can keep this and change the others to dashes, agreed.

src/passes/analysis/verilog.cpp
rsetaluri: Yeah on second thought, I think it's fine to do as a pass.

src/passes/analysis/verilog.cpp
rsetaluri: ugh yeah I see, really the passes should have a `finalize()` hook, but doesn't look like it's in the infra right now. problem is that `writeToFiles()` also needs this logic. I would suggest encapsulating this as follows, (assuming `b ool prefixAdded = False;` is in the class declaration in the hpp)

```cpp
void Passes::Verilog::addPrefix() {
    if (prefixAdded || module_name_prefix == "") return;
    // Renaming logic here...
    for (auto& module : modules) {
        ...
    }
    prefixAdded = True;
}

 void Passes::Verilog::writeToStream(std::ostream& os) {
    addPrefix();
    // Rest of logic here...
}

 void Passes::Verilog::writeToFiles(...) {
    addPrefix();
    // Rest of logic here...
}
```

src/passes/analysis/verilog.cpp
```diff
@@ -1594,14 +1600,60 @@ bool Passes::Verilog::runOnInstanceGraphNode(InstanceGraphNode& node) {
...
             "Module `" + module->getName() + "` defined externally")
             .toString()
        << std::endl;
   }
-  for (auto& module : modules) { os << module.second->toString() << std::endl; }
+  if (this->module_name_prefix != "") {
+    std::set<std::string> renamed_modules;
+    for (auto& module : modules) {
+      bool added = add_prefix(module.second, this->module_name_prefix);
+      if (added) { renamed_modules.insert(module.first); }
```
rsetaluri, line 1647: nit
```suggestion
      if (added) renamed_modules.insert(module.first);
```

rsetaluri (approved):
a few nits, and then comment re: encapsulating add prefix outer loop. good to go after that.

PR #992, opened 2021-03-05
Adds support for outputs inside a compile guard

leonardt:
By default, we drive outputs to be 0 inside the else clause (should work
for arbitrary ports since types have been flattened).

This allows for "stubs" to be inserted when a compile_guard is not
active.

This feature is simpler than supporting a general output handler in the
compile guard, since then we'd need to support multiple drivers from
different compile guard cases which is tricky since in magma/coreir we
only allow a single driver for a port.  This covers the use case at FB
without having to dive into the details of general output handling.


src/passes/analysis/verilog.cpp
```diff
@@ -1368,18 +1368,34 @@ Passes::Verilog::compileModuleBody(
...
-        body.push_back(std::move(statement));
+        std::vector<std::unique_ptr<vAST::StructuralStatement>> true_body;
+        true_body.push_back(std::move(statement));
         std::string type = metadata["compile_guard"]["type"];
+
+        std::vector<std::unique_ptr<vAST::StructuralStatement>> else_body;
+        RecordType* record_type = cast<RecordType>(instance_module->getType());
+        for (auto field : record_type->getFields()) {
+          Type* field_type = record_type->getRecord().at(field);
+          std::string wire_name = instance.first + "_" + field;
```
rsetaluri, line 1379: This seems extremely fragile...

rsetaluri (commented):
I agree that to unblock for the output-driving use case this logic works, but it looks very fragile and hard to maintain.

I actually think a pass at the magma level to handle this more robustly and generally would not be too crazy. Here is what I envision:

- When processing outputs of a compile-guarded instance, collect the drivees and add new output ports to the compile-guarded circuit, but do not yet re-wire them. (basically up to `magma/compile_guard.py:95` in magma#935.) We need a data structure that basically maps the `external` to `drivee`.
- Run a CircuitPass over all circuit definitions that does the following:
    * Find all instances which are compile-guarded. If there are none, return.
    * Find all external `drivee` ports which need to be wired to. If there are none, return.
    * For all external `drivee` ports which appear only once, wire them to their corresponding `external` port (we know this from the data structure kept in the first bullet).
    * For all "multiply" (`N`) driven `drivee` ports, create an intermediate "adapter" circuit, which has as its interface, `N` inputs, and 1 output. Wire the output to the `drivee` and each of the multiple drivers to the `N` inputs. Let's come back to what this circuit is.

Now everything is structurally sound; the drivee ports only have one driver, regardless of how many times you tried to wire to it from *different* compile-guarded blocks.

Now is the question of what the "adapter" is. I think we can create a new CoreIR generator (maybe in mantle library, or a new verilog_utils library). This generator has generator parameters: `type: CoreIRType` and `conditions: List[String]`. The interface is exactly what I described above, and the impl looks like:

```verilog
`ifdef {conditions[0]}
    assign O = I0;
`elsif {conditions[1]}
    assign O = I1;
...
`endif
```

This would take a *bit* of work but conceptually isn't too crazy. I'm ok doing this change now and keeping track of the fragility, and shooting to replace it with something more robust fairly soon.

src/passes/analysis/verilog.cpp
leonardt: This is how output wires are generated by coreir, can you offer a better less fragile solution? The point of having a separate inline pass is that at this point we can assume that the output wires have this name without having to maintain them since the inlining pass handles that.  I'm open to better suggestions but I'm not sure if I see one

leonardt: Read your full review comment after this one, seems like you're just marking place where something could be improved.  We should open an issue to track this and reference the comments here.

rsetaluri: Yeah, it's probably not worth it to make it more robust for this version. But I think there's a solution that basically solves "all" the problems. My main concern re: fragility here was how "guaranteed" the names for output wires are `instance.first + "_" + field`. E.g. if one line changed somewhere else in how that name was generated, would this break, but the rest of code generation would work (i.e. is not dependent on how those wire names are gen'd)?

rsetaluri: I think opening a new issue + just a comment here on the explanation of this logic + what it does and doesn't work for would suffice for now (e.g. "handles single driver of output, but not multiple drivers from different conditional blocks etc."

rsetaluri (approved):

PR #997, opened 2021-04-01
Instrument inline symbol table

rsetaluri:


src/ir/moduledef.cpp
```diff
@@ -214,37 +215,33 @@ Instance* ModuleDef::getInstancesIterNext(Instance* instance) {
...
-
 Instance* ModuleDef::addInstance(string instname, Module* m, Values modargs) {
   ASSERT(instances.count(instname) == 0, instname + " already an instance");
   Instance* inst = new Instance(this, instname, m, modargs);
   instances[instname] = inst;
-
   appendInstanceToIter(inst);
 
+  // Log new instance for symbol table.
+  {
```
rdaly525, line 225: Nit: shouldn't this be one more indent level

src/passes/analysis/verilog.cpp
```diff
@@ -1411,10 +1411,10 @@ Passes::Verilog::compileModuleBody(
           throw std::runtime_error("Unexpected compile_guard type: " + type);
         }
       }
-      this->getSymbolTable()->setInstanceName(
-          definition->getModule()->getName(),
-          instance.second->getInstname(),
-          instance_name);
+      // this->getSymbolTable()->setInstanceName(
```
rdaly525, line 1414: Should remove

tests/gtest/test_flattentypes.cpp
```diff
@@ -20,6 +21,42 @@ TEST(FlattenTypesTests, TestFlattenTypesPreserveNDArray) {
   deleteContext(c);
 }
 
+
+//void check_port(SymbolTableInterface* sym, std::string key, std::string val) {
```
rdaly525, line 25: Should remove

tests/gtest/test_inline_instance.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+  std::string flattentypes_str = "flattentypes";
+  ctx->runPasses({"flattentypes", "verilog"}, {});
+
+  table = ctx->getPassManager()->getSymbolTable();
+  table->getLogger()->finalize();
+  // Check module names (sanity check).
+  EXPECT_EQ(table->getModuleName("M1"), "M1");
+  // This won't work since there are no longer instances of M2:
+  //   EXPECT_EQ(table->getModuleName("M2"), "M2");
```
rdaly525, line 78: Why does the symbol table remove M2 if an instance of M2 is inlined? what if there are other instances of M2 in other modules?

rdaly525 (approved):
Looks good, minor nits

src/ir/moduledef.cpp
rsetaluri: When I scope things I don't like that extra level of indentation. I'll just get rid of the scoping.

src/passes/analysis/verilog.cpp
rsetaluri: I should really insert a `"RENAME"` here but we'll have to deal with that in finalization later. Removing for now and keeping a TODO comment.

tests/gtest/test_flattentypes.cpp
rsetaluri: done

tests/gtest/test_inline_instance.cpp
rsetaluri: then it won't be removed. its just because the module names are only inserted into the symbol table during verilog generation which is an instance graph pass. if a module is "unused" in that, then it won't appear in the symbol table.

src/ir/inline.cpp
```diff
@@ -231,7 +235,13 @@ bool inlineInstance(Instance* inst) {
   // I will be inlining defInline into def
   // Making a copy because i want to modify it first without modifying all of
   // the other instnaces of modInline
-  ModuleDef* defInline = modInline->getDef()->copy();
+  auto defInline = [&]() {
```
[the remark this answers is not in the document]
rsetaluri: doesn't need to be. simplifying.

src/ir/inline.cpp
```diff
@@ -326,4 +343,47 @@ bool inlineInstance(Instance* inst) {
   return true;
 }
 
+}  // namespace
+
+Instance* addPassthrough(Wireable* w, string instname) {
```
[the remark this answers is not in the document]
rsetaluri: not strictly needed but can be a nice pattern if you have to do stuff outside like we do for inline, so just copied that pattern here. can revert.

PR #1001, opened 2021-04-08
Use module longmame for symbol table

rsetaluri:
Using a module's longname allows us to include generators and circuits
across namespaces.


rdaly525 (approved):
As something to watch out for, I think anything in the 'global' namespace is special cased for the longname (it does not include the word 'global')

rsetaluri:
@rdaly525 yup, I think that is assumed by magma side as well.

PR #1002, opened 2021-04-08
adds get function for instance metadata

rdaly525:


src/coreir-c/coreir-c.cpp
```diff
@@ -455,6 +455,17 @@ const char* COREModuleGetMetaData(COREModule* core_mod) {
...
 
+const char* COREInstanceGetMetaData(COREWireable* core_inst) {
+  auto w = rcast<Wireable*>(core_inst);
+  ASSERT(isa<Instance>(w), "Wireable needs to be an instnace");
+  Instance* inst = cast<Instance>(w);
+  string istr = inst->getMetaData().dump();
+  std::size_t len = istr.size() + 1;
+  char* cstr = (char*)malloc(len);
+  strcpy(cstr, istr.c_str());
+  return cstr;
```
rsetaluri, line 466: let's use context scratchpad: https://github.com/rdaly525/coreir/blob/1e30099a7c5a08f741777e37821d96e73e9d82d3/src/coreir-c/types-c.cpp#L74-L81

rsetaluri (approved):

src/coreir-c/coreir-c.cpp
rsetaluri, line 466:
```suggestion
const char* COREInstanceGetMetaData(COREWireable* core_inst) {
  auto w = rcast<Wireable*>(core_inst);
  ASSERT(isa<Instance>(w), "Wireable needs to be an instnace");
  Instance* inst = cast<Instance>(w);
  std::string metadata = inst->getMetaData().dump();
  auto copy = static_cast<char*>(
      w->getContext()->getScratchMemory(metadata.size() + 1));
  strcpy(copy, metadata.c_str());
  return copy;
```

rsetaluri (approved):

PR #1012, opened 2021-05-20
Update coreir-c.cpp

rsetaluri:


include/coreir-c/coreir.h
```diff
@@ -215,6 +215,7 @@ extern void COREModuleAddMetaDataStr(
   char* value);
 
 // This returns a malloc'd string that should be freed by the user
+extern const char* COREGeneratorGetMetaData(COREGenerator* module);
 extern const char* COREModuleGetMetaData(COREModule* module);
 extern const char* COREInstanceGetMetaData(COREWireable* module);
```
rsetaluri, line 220: Can we collapse these 3 into a single function since they all inherit from the `Metadata` base class?

rdaly525: It seems like that should be possible. Specifying the types can get a little annoying though. There is no current method to 'cast' from a COREModule* to a COREMetadata*

src/coreir-c/coreir-c.cpp
```diff
@@ -455,6 +455,15 @@ const char* COREModuleGetMetaData(COREModule* core_mod) {
   return cstr;
 }
 
+const char* COREModuleGetMetaData(COREGenerator* core_gen) {
+  Generator* gen = rcast<Generator*>(core_gen);
+  string mstr = gen->getMetaData().dump();
+  std::size_t len = mstr.size() + 1;
+  char* cstr = (char*)malloc(len);
```
rdaly525, line 462: How does this get freed? Looks like the instance metadata API allocates from context's scratchmemory.

src/coreir-c/coreir-c.cpp
```diff
@@ -455,6 +455,15 @@ const char* COREModuleGetMetaData(COREModule* core_mod) {
   return cstr;
 }
 
+const char* COREModuleGetMetaData(COREGenerator* core_gen) {
```
rdaly525, line 458: Where is the Generator metadata implementations?

rdaly525 (changes requested):
See comments

rsetaluri:
@rdaly525 closing this for now since I no longer need this.

Issue #941, opened 2020-08-17
Add a Custom Name for CoreIR Generator When Generate Verilog


joyliu37:
I am using coreIR to generate a memory placeholder in RTL for further power analysis. 
I want the RTL module which coreIR generator generated have the specific name matching with RTL implementation. 
Currently the RTL module name is automatic generated from the `gen_arg`. For instance, my [generated verilog](namehttps://github.com/dillonhuff/clockwork/blob/ef2d82ee46539c682c0f04263ece4b821def4da0/conv33_naive_compute.v#L118) have this wield name resulted from the type. 
Could we have a function to set custom name for RTL generation?

joyliu37:
@rdaly525 Any update on this feature? I found the name is too long for some circumstances.

leonardt:
There's a metadata field called `verilog_name` that can be used to override the default name generation in coreir.  Perhaps inside your generator logic, when creating a module, you could try setting this field and seeing if it works for you?

Here's an example from the ice40 generator: https://github.com/rdaly525/coreir/blob/7a8106264314fb2b1e89f830af447ed40e893fed/src/libs/ice40.cpp#L19-L20 (we need the generated verilog module to have a specific name for synthesis)

joyliu37:
Thanks Lenny. It does not work for me. In the verilog generated from coreIR, the name of that module is not override by the `verilog_name`. Do I need to change other places? 
There are other data field in my metadata, is that matters?

leonardt:
Hmm that's odd, any chance you can link me to a branch or something where the code is written? Perhaps we're setting the metadata in the wrong place, let me review the code generation logic for generators, maybe it's different and ignoring the metadata field (the example I provided was just a module not a generator, so maybe it only works for modules)

joyliu37:
Yeah, I think the difference between module and generator may affect. I print out the meta-data in that instance and it's empty. 
I push my code in the `unsharp_fix` branch in clockwork, and here is the part I encode the name into the generator module. https://github.com/dillonhuff/clockwork/blob/793ed8efadcfdb2d4511f06c9f406a9031b1fecf/ubuffer.cpp#L2366 

To quickly reproduce the result I see you could attach my docker on kiwi and run the command in clockwork repo
```
docker attach joeyliu-quizzical_wilbur
:/aha/clockwork# ./rebuild_set_test.sh lake-tests
```

rdaly525:
[...]

PR #1013, opened 2021-06-25
add verilog customized name for generator

joyliu37:
I made some changes to the verilog analysis passes and it work for me for the issue. 
https://github.com/rdaly525/coreir/issues/941

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -1135,6 +1135,9 @@ Passes::Verilog::compileModuleBody(
         json verilog_json = instance_module->getGenerator()
                               ->getMetaData()["verilog"];
         module_name = make_name(instance_module->getName(), verilog_json);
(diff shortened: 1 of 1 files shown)


leonardt:
It seems like this is the code we should be trying to use: https://github.com/rdaly525/coreir/pull/1013/files#diff-1d13ccebd6fd308ba92fd6d7889462cc6b4a370afdbc5edb561dfe2ddab84781R1147-R1154 but because they metadata is lost when you save to JSON it's not working.  Let me investigate the options there, perhaps we should save that metadata to be persistent across invocations.  However, it's likely that you should also just be able to save the verilog file directly from the runtime which would avoid that issue.

rdaly525:
@joyliu37 if you use serializeToFile(...) instead of saveToFile(...) All the metadata and dependencies will get saved in the one verilog file.

PR #1015, opened 2021-07-01
Update coreir_to_dot to output op instead of names

jeffsetter:
Writes coreir operator to dot file rather than instance name


rdaly525 (changes requested):
Please remove all the commented code. Thanks!

I just emailed travis regarding the travis issues so I should hopefully be able to merge soon

rdaly525 (approved):

PR #1016, opened 2021-07-22
Fixes bug in verilog name generation logic

### LINKED ISSUE #941: Add a Custom Name for CoreIR Generator When Generate Verilog

rdaly525:
Fixes #941

### DIFF
--- a/src/passes/analysis/verilog.cpp
+++ b/src/passes/analysis/verilog.cpp
@@ -1130,12 +1130,9 @@ Passes::Verilog::compileModuleBody(
   for (auto instance : instances) {
     Module* instance_module = instance.second->getModuleRef();
     std::string module_name = instance_module->getLongName();
-    if (instance_module->isGenerated()) {
-      if (instance_module->getGenerator()->getMetaData().count("verilog") > 0) {
-        json verilog_json = instance_module->getGenerator()
-                              ->getMetaData()["verilog"];
-        module_name = make_name(instance_module->getName(), verilog_json);
-      }
+    if (instance_module->isGenerated() && instance_module->getGenerator()->getMetaData().count("verilog") > 0) {
+      json verilog_json = instance_module->getGenerator()->getMetaData()["verilog"];
+      module_name = make_name(instance_module->getName(), verilog_json);
     }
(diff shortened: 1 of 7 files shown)


joyliu37 (commented):
Looks good to me. So I need to use `serializeToFile` to dump the generator information?

joyliu37 (approved):

src/passes/analysis/verilog.cpp
```diff
@@ -1130,12 +1130,9 @@ Passes::Verilog::compileModuleBody(
   for (auto instance : instances) {
     Module* instance_module = instance.second->getModuleRef();
     std::string module_name = instance_module->getLongName();
-    if (instance_module->isGenerated()) {
-      if (instance_module->getGenerator()->getMetaData().count("verilog") > 0) {
-        json verilog_json = instance_module->getGenerator()
-                              ->getMetaData()["verilog"];
-        module_name = make_name(instance_module->getName(), verilog_json);
-      }
+    if (instance_module->isGenerated() && instance_module->getGenerator()->getMetaData().count("verilog") > 0) {
```
joyliu37, line 1133: I change this line of code into
```
if (instance_module->isGenerated() && instance_module->getMetaData().count("verilog") > 0)
```

joyliu37 (changes requested):

src/passes/analysis/verilog.cpp
leonardt: @rdaly525 I think this means you need to check two separate cases here:
1. the instance_module has `"verilog"` metadata
2. the instance_module->getGenerator() has `"verilog"` metadata

leonardt (approved):
Change looks fine, but I think there's one extra case needed for Joey's use case

PR #1023, opened 2021-11-29
changes master to main for gtest cmake dep

rdaly525:


rdaly525:
gtest changed master to main

rdaly525:
the OSX build is timing out before it even gets to building the coreir repo.

rdaly525:
Fixes #1023 #1024. Since this works in linux and is breaking the aha build, I am merging.

